PR #214, opened 2026-04-22
Add tbl hierarchical incidence rate

### LINKED ISSUE #211: Add tbl_hierarchical_incidence_rate()
### Feature description

Currently template code for AET02_ADJ is very long and uses repeated execution of the same code for calculating indicence rate. 
The best option would be to abstract this calculations in a new function in crane, which handles the calculations.
This will shorten the AET02_ADJ template code and improve user experience.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [x] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [x] I agree to follow this project's Security Policy.

jszczypinski:
* Style this entry in a way that can be copied directly into `NEWS.md`. (#<issue number>, @<username>)

Provide more detail here as needed.

**Reference GitHub issue associated with pull request.** _e.g., 'closes #<issue number>'_

Closes #211 

- [ ] Request a reviewer

- [ ] Approve Pull Request


R/tbl_hierarchical_incidence_rate.R
```diff
@@ -0,0 +1,377 @@
...
+  check_data_frame(denominator)
+  check_numeric(n_person_time)
+  check_scalar(n_person_time)
+  check_string(unit_label)
+  check_numeric(conf.level)
+  check_scalar(conf.level)
+  check_string(conf.type)
+  check_numeric(digits)
+  check_scalar(digits)
+
```
llrs-roche: There are no checks for some arguments: `variables`, `by`, `id`, `start_date`, `end_date`, `event_date` and `label`. The checks are cursory, they check class but not length or dimension in most cases, or at least ont initially, as some are scattered below (`check_length(variables, ...)`)

R/tbl_hierarchical_incidence_rate.R
```diff
@@ -0,0 +1,377 @@
...
+  tbl_final <- c(list(tbl_base), tbls_rates) |>
+    gtsummary::tbl_merge(tab_spanner = FALSE) |>
+    gtsummary::modify_table_body(\(x) {
+      col_order <- sort(names(dplyr::select(x, gtsummary::all_stat_cols())))
+      dplyr::relocate(x, dplyr::all_of(col_order), .after = "label")
+    }) |>
+    gtsummary::remove_footnote_header(tidyselect::everything())
+
+  tbl_final$call_list <- list(tbl_hierarchical_incidence_rate = match.call())
+  tbl_final$inputs <- as.list(environment())
```
llrs-roche: By saving here the environment we get 25 elements instead of 15 if we do it at the top. Thus including more than just the input (all the internal variables created).  Move this line to line 129 or above.

R/tbl_hierarchical_incidence_rate.R
```diff
@@ -0,0 +1,377 @@
...
+  tbls_rates <- lapply(names(tbl_stat_labels), function(stat) {
+    list(
+      gtsummary::tbl_ard_summary(
+        ard_overall,
+        by = dplyr::all_of(by), statistic = ~stat
+      ) |>
+        gtsummary::modify_table_body(~ .x |> dplyr::mutate(
+          row_type = "level",
+          var_label = NA,
+          label = .env$overall_label,
```
llrs-roche: Why do we use `.env$overall_label` instead of `overall_label`? 

Starting from a list of tables and staking them seems like a challenging situation. Can't the table be calculated from a single call? `gtsummary::tbl_ard_summary(statistic)` accepts a list.

R/tbl_hierarchical_incidence_rate.R
```diff
@@ -0,0 +1,377 @@
...
+    overall_row = TRUE
+  ) |>
+    gtsummary::sort_hierarchical() |>
+    gtsummary::modify_header(
+      gtsummary::all_stat_cols() ~ "No. of\nParticipants\nwith AE (%)"
+    ) |>
+    gtsummary::modify_table_body(~ .x |> dplyr::mutate(
+      label = ifelse(
+        dplyr::row_number() == 1,
+        .env$overall_label,
```
llrs-roche: Why do we use `.env$overall_label` instead of `overall_label`? Similar to what's about `.data$label`  instead of just `label`

R/tbl_hierarchical_incidence_rate.R
```diff
@@ -0,0 +1,377 @@
...
+    denominator = denominator,
+    label = label,
+    overall_row = TRUE
+  ) |>
+    gtsummary::sort_hierarchical() |>
+    gtsummary::modify_header(
+      gtsummary::all_stat_cols() ~ "No. of\nParticipants\nwith AE (%)"
+    ) |>
+    gtsummary::modify_table_body(~ .x |> dplyr::mutate(
+      label = ifelse(
```
llrs-roche: Are you sure you want to iterate over all rows with the initial content?  In other conditions you `if_else` is used:
```suggestion
      label = dplyr::if_else(
```

R/tbl_hierarchical_incidence_rate.R
```diff
@@ -0,0 +1,377 @@
...
+                                by,
+                                start_date,
+                                end_date,
+                                event_date,
+                                strata_vars = NULL,
+                                n_person_time = 100,
+                                unit_label = "years",
+                                conf.level = 0.95,
+                                conf.type = "normal",
+                                digits = 2) {
```
llrs-roche: There are three different calls to `.ard_incidence_rate` with only changes on `strata_vars` and how the columns are named `variable(_level)?` . Do we need such a big function with that many parameters? It looks like a simpler one could work too. That could clean/avoid more refactoring after calling it as it already gets `by` to do the renaming inside the function if needed.

R/tbl_hierarchical_incidence_rate.R
```diff
@@ -0,0 +1,377 @@
...
+  # Data manipulation ----------------------------------------------------------
+  if (!is.null(strata_vars) && length(strata_vars) > 0) {
+    # expand the denominator to include all patients x events
```
llrs-roche: I think that instead of unique and join with dummy data it can be simplified with `dplyr::expand_grid()` or `expand.grid()`

R/tbl_hierarchical_incidence_rate.R
```diff
@@ -0,0 +1,377 @@
...
+    # to calculate time to event also for patients without AE reported
```
llrs-roche: I don't see any calculation on this code block about time to event

R/tbl_hierarchical_incidence_rate.R
```diff
@@ -0,0 +1,377 @@
...
+  } else {
+    expanded_denom <- denominator
+  }
+
+  grp_vars <- c(id, by, strata_vars)
+
+  # Summarize events -----------------------------------------------------------
+  safe_min_date <- function(x) {
+    x_char <- as.character(x)
+    x_char[x_char == "" | x_char == "NA"] <- NA_character_
```
llrs-roche: Not sure but on previous NEST tables there was an option to set a default string for NA. Is this still working? In that case you might need to check the options or something else to replace by NA.

In any case safe_min_date seems an utility function that could be on `utils.R` file. as it doesn't need to be inside `.ard_incidence_rate()`.

R/tbl_hierarchical_incidence_rate.R
```diff
@@ -0,0 +1,377 @@
...
+  # Join counts onto the expanded grid and calculate Time-to-Event -------------
+  merged_data <- expanded_denom |>
+    dplyr::left_join(ae_counts, by = grp_vars) |>
+    dplyr::mutate(
+      count = dplyr::coalesce(.data$count, 0L),
+      # Parse start and end dates once to keep the if_else clean
+      start_dt = as.Date(
+        substr(as.character(.data[[start_date]]), 1, 10),
+        format = "%Y-%m-%d"
+      ),
```
llrs-roche: This snippet is repeated in `safe_min_date` and below. Maybe can be an utility too?

```r
as_date <- function(x) {
    as.Date(substr(as.character(x), 1, 10), format = "%Y-%m-%d")
}
```

R/tbl_hierarchical_incidence_rate.R
```diff
@@ -0,0 +1,377 @@
...
+      time_var = dplyr::if_else(
+        # Check for valid dates that occur ON or AFTER the start date
+        .data$count > 0 &
+          !is.na(.data$ae_onset) &
+          .data$ae_onset >= .data$start_dt,
+        # Valid AE Date: Onset date - Start date + 1
+        as.numeric(
+          difftime(.data$ae_onset, .data$start_dt, units = "days") + 1
+        ) / 365.25,
```
llrs-roche: I am not sure what is the convention for years (365.25 or just 365) but it might be better to convert it to dates and then calculate difference on dates not on times.

R/tbl_hierarchical_incidence_rate.R
```diff
@@ -0,0 +1,377 @@
...
+    variables = {{ variables }},
+    by = {{ by }},
+    id = {{ id }},
+    event_date = {{ event_date }}
+  )
+  cards::process_selectors(
+    denominator,
+    start_date = {{ start_date }},
+    end_date = {{ end_date }}
+  )
```
llrs-roche, line 128: Just a side not for me to understand how crane works: I'm not sure I understood what is the purpose of `process_selector()` but there are two calls to it instead of just one. I know that data changed but there is no comment and the help page of card don't explain why and how.

R/tbl_hierarchical_incidence_rate.R
```diff
@@ -0,0 +1,377 @@
...
+    cardx::ard_total_n(denominator)
+  )
+
+  # Format the header labels based on the unit_label
+  pt_abbr <- switch(tolower(unit_label),
+    "years" = "PY",
+    "months" = "PM",
+    "days" = "PD",
+    "weeks" = "PW",
+    "time" = "PT",
```
llrs-roche, line 217: Where do these labels come from? I don't seem to find them anywhere else on the package.

R/tbl_hierarchical_incidence_rate.R
```diff
@@ -0,0 +1,377 @@
...
+      dplyr::relocate(x, dplyr::all_of(col_order), .after = "label")
+    }) |>
+    gtsummary::remove_footnote_header(tidyselect::everything())
+
+  tbl_final$call_list <- list(tbl_hierarchical_incidence_rate = match.call())
+  tbl_final$inputs <- as.list(environment())
+
+  tbl_final |> structure(
+    class = c("tbl_hierarchical_incidence_rate", "gtsummary")
+  )
```
llrs-roche: This replaces the `tbl_merge` class by this class (I don't see any method with `methods(class = "tbl_merge")`). Also structure can be used to create any object. I think it is better to use `class()<-` than use `structure()`.

tests/testthat/test-tbl_hierarchical_incidence_rate.R
```diff
@@ -0,0 +1,145 @@
...
+      unit_label = "decades"
+    )
+
+    header_df <- tbl_custom_unit$table_styling$header
+    expect_true(any(grepl("Person-Decades", header_df$label)))
+  }
+)
+
+test_that(
+  "tbl_hierarchical_incidence_rate fails safely on bad inputs",
```
llrs-roche: It only covers three parameters not all the 14 ones

tests/testthat/test-tbl_hierarchical_incidence_rate.R
```diff
@@ -0,0 +1,145 @@
...
+      variables = c(AESOC, AEDECOD),
+      start_date = TRTSDT,
+      end_date = TRTEDT,
+      event_date = AESTDTC,
+      # Forces the switch fallback: paste0("Person-", "Decades")
+      unit_label = "decades"
+    )
+
+    header_df <- tbl_custom_unit$table_styling$header
+    expect_true(any(grepl("Person-Decades", header_df$label)))
```
llrs-roche: The input is with lower case but the output is with the first letter in upper case. I know this is intended, but there is no explanation of it on the help page or examples

tests/testthat/test-tbl_hierarchical_incidence_rate.R
```diff
@@ -0,0 +1,145 @@
...
+    "2023-06-01", # Standard clean date
+    "2023-03-15T12:30", # ISO 8601 with time
+    "2022-12-01", # Prior to start date (calculates healthy time)
+    "", # Blank string
+    "NA" # Literal NA string
+  )
+)
+
+test_that(
+  "tbl_hierarchical_incidence_rate executes standard usage with `by`",
```
llrs-roche: Not sure if standard but the function only requires 3 arguments and I don't see any test with only those three: data, denominator, variables.  I would appreciate if we have one test with only those.

tests/testthat/test-tbl_hierarchical_incidence_rate.R
```diff
@@ -0,0 +1,145 @@
+# tests/testthat/test-tbl_hierarchical_incidence_rate.R
+
+library(testthat)
+library(dplyr)
+
```
llrs-roche: This is not needed.

llrs-roche (changes requested):
Overall I think it is great addition that simplifies the table manipulation a lot on a common case. 

However, it is my first time reviewing a feature on crane and I am not familiar with conventions and utilities available, so I left several comments on this direction.
I also think that the code could be simpler and easier to read but I might be wrong. 
For example, could it be used for other tables with adverse events?

There are some test that cover 100% but I am not sure they cover what needs to be covered, so I left some comments there too. 
Feel free to challenge my questions or disagree with my comments!

R/tbl_hierarchical_incidence_rate.R
Melkiades: This is a very good point. I always thought {tern} was a bit "hurried" on this as it uses the above approximation and others for AE

Melkiades: We might want to copy {tern} 1-1 for numerical results while adding a recognizable comment. Then, we properly format the calculations in one PR that will inevitably diverge from {tern} results?

R/tbl_hierarchical_incidence_rate.R
Melkiades: They allow for tidyr selectors to be processed into a normal character vector. This is useful because you have all error messaging there and it is always working against a specific data.frame (hence the two calls)

R/tbl_hierarchical_incidence_rate.R
jszczypinski: Unfortunately, I think we have to use such a big function. The function needs data, denominator, id, by, dates and strata_vars to create the data needed for `ard_incidence_rate` to run the calculations. On the other hand, `ard_incidence_rate` needs all other parameters and if we want for users to be able to control these parameters then we have to pass them all. Nevertheless, I think it will be better to split the function in two: 1) to prepare data 2) to make calculations. Also I moved the rename function inside the helper.

R/tbl_hierarchical_incidence_rate.R
jszczypinski: I use these so that R CMD Check can correctly evaluate. The `.data$label` inform R CMD Check that the label variable belongs to the dplyr pipeline, while using `env$overall_label` tells it that it belong to the function environment. In short it handles unwanted/unexpected data masking. More here 
https://tinyheero.github.io/2020/03/01/use-data-env-pronouns-tidyverse.html

R/tbl_hierarchical_incidence_rate.R
jszczypinski: You are right, changed that do `dplyr::cross_join()` - it does the same thing as `expand_grid()`. `expand_grid()` is from tidyr

R/tbl_hierarchical_incidence_rate.R
llrs-roche: Yes but he overall_label exists on the local environment L111 to L113 and I think we want to use that.

R/tbl_hierarchical_incidence_rate.R
llrs-roche: If we want to pass arguments it would be safe to use `...` but splitting in two is also a good idea, it will make easier to reuse them in the future.

R/tbl_hierarchical_incidence_rate.R
llrs-roche: It seems tern uses 365.24: https://github.com/insightsengineering/tern/blob/14bd582665967f7a5bea25c3e15ccb271b190eb0/R/incidence_rate.R#L91 (typo?, correction for some non leap years?)

I think this is critical to be documented and constant across the framework and exposed/shared for the users to know. It might require some work on tern but I think this is important. The path proposed  seems good.

R/tbl_hierarchical_incidence_rate.R
jszczypinski: It would be great to use a single call. Yet, if we would calculate the tables from the single call using list of statistics then we will get all statistics stacked vertically, while the we need a single column for each statistic stacked horizontally. That's why I use a loop here

R/tbl_hierarchical_incidence_rate.R
jszczypinski: Done

R/tbl_hierarchical_incidence_rate.R
jszczypinski: Removed that comment

R/tbl_hierarchical_incidence_rate.R
jszczypinski: Changed it to  `class(tbl_final) <- c("tbl_hierarchical_incidence_rate", class(tbl_final))`
This way `tbl_merge` will not be replaced

R/tbl_hierarchical_incidence_rate.R
jszczypinski: This is standard naming convention for Person-years (PY), Person-months (PM), Person-days (PD) and so on.

R/tbl_hierarchical_incidence_rate.R
jszczypinski: This is a standard step used in crane functions. `process_selectors()` takes these dataframes and arguments, which are tidyselectors (USUBJID, TRTSDT) and saves them to fhe calling environment as character vectors.

jszczypinski: I have added some comments

R/tbl_hierarchical_incidence_rate.R
jszczypinski: Check scalar, check_string and so on check the length as well, by default length = 1L.
The checks are scattered cause some of the arguments can be tested only after `process_selectors()` are executed.
I have added some more checks have full coverage

R/tbl_hierarchical_incidence_rate.R
jszczypinski: Done

R/tbl_hierarchical_incidence_rate.R
jszczypinski: Done

R/tbl_hierarchical_incidence_rate.R
jszczypinski: Since I do operate on these arguments inside the helper functions, then I would have to save them using `args <- list(...)` or something similar and then use them as `args$data`, `args$denom` and so on. I am not sure if it will improve readibility

R/tbl_hierarchical_incidence_rate.R
jszczypinski: I moved it outside `.ard_incidence_rate()`. For now it is only used in `tbl_hierarchical_incidence_rate()` so we will not move it to `utils.R`

R/tbl_hierarchical_incidence_rate.R
jszczypinski: Same as above. Made it a helper function, but for now it should stay in `tbl_hierarchical_incidence_rate()`

R/tbl_hierarchical_incidence_rate.R
jszczypinski: You are totally right, but the `env$overall_label` is using `overall_label` from lines 111-113. `.env` will point to a variable from the local execution environment

tests/testthat/test-tbl_hierarchical_incidence_rate.R
jszczypinski: Removed

R/tbl_hierarchical_incidence_rate.R
jszczypinski: Ther {tern} probably uses a rounded down version of `365.2425`. 
I will change it to 365.24 to match {tern} and do the calculations on Dates.
Should I open the issue regarding updating the dates calculations in {crane}?

tests/testthat/test-tbl_hierarchical_incidence_rate.R
jszczypinski: Good call, I have added the test

tests/testthat/test-tbl_hierarchical_incidence_rate.R
jszczypinski: Added a comment about that. Also I have updated the documentation of the function

tests/testthat/test-tbl_hierarchical_incidence_rate.R
jszczypinski: I added tests for all 14

tests/testthat/test-tbl_hierarchical_incidence_rate.R
```diff
@@ -0,0 +1,170 @@
...
```
llrs-roche, line 3: The library calls are not needed.

llrs-roche (approved):
Looking good ant tests are passing locally

PR #232, opened 2026-05-10
Add `tbl_ancova()` for ANCOVA tables with treatment-vs-control contrasts

### LINKED ISSUE #6: New `tbl_*()` for LS means/ANCOVA
This table seems to appear a few times in the catalog. Probably should just make a function for it.

https://insightsengineering.github.io/tlg-catalog/stable/tables/efficacy/aovt01.html
Image

Melkiades:
* `tbl_ancova()` builds ANCOVA tables with adjusted means and treatment-vs-control contrasts using `emmeans`. (#6, @garolind)

The function fits a model via `cardx::construct_model()`, computes LS means with `emmeans::emmeans()`, and obtains contrasts with `emmeans::contrast(method = "trt.vs.ctrl")`. Each non-reference group is compared to the reference group only (not all pairwise). The output is a gtsummary table showing n, adjusted mean, difference in adjusted means, CI, and p-value per treatment group.

Key design decisions:
- Uses `trt.vs.ctrl` instead of `pairwise` contrasts to match the standard ANCOVA table layout where each treatment is compared to a single control group.
- Accepts a `ref_group` argument to specify the control arm.
- Accepts a `denominator` argument for header Ns from a subject-level dataset.
- Accepts `conf.level` for configurable CI width (default 95%).
- Accepts `method`, `method.args`, `package` for flexible model specification.

**Reference GitHub issue associated with pull request.** closes #6

- [ ] Request a reviewer


R/tbl_ancova.R
```diff
@@ -0,0 +1,301 @@
...
+#' - Confidence interval for the difference
+#' - p-value for the difference
+#'
+#' The model is fit via [`cardx::construct_model()`] and least-squares
+#' estimates are obtained with the `emmeans` package. Contrasts use
+#' `emmeans::contrast(method = "trt.vs.ctrl")` so that each non-reference
+#' group is compared to the reference group only.
+#'
+#' @param data (`data.frame`)\cr
+#'   Analysis data set, typically one parameter/visit subset of an ADaM BDS.
```
Melkiades: BDS?

R/tbl_ancova.R
```diff
@@ -0,0 +1,301 @@
...
+#'   [`emmeans::contrast()`]. Common choices: `"none"` (no adjustment,
+#'   default), `"dunnett"` (Dunnett's method), `"bonferroni"`, `"tukey"`.
+#'   See [`emmeans::summary.emmGrid()`] for all options.
+#' @param denominator (`data.frame`)\cr
+#'   Optional data frame used to compute the header Ns (typically `ADSL`).
+#'   When supplied, the column headers show `(N = <count>)` from this data
+#'   frame rather than from `data`.
+#'
+#' @return A `'gtsummary'` table of class `c("tbl_ancova", "gtsummary")`.
+#' @name tbl_ancova
```
Melkiades, line 54: This to move above the export declaration (I prefer to have it clear there)

R/tbl_ancova.R
```diff
@@ -0,0 +1,301 @@
...
+#'
+#' @examplesIf identical(Sys.getenv("NOT_CRAN"), "true") || identical(Sys.getenv("IN_PKGDOWN"), "true")
```
Melkiades: This should happen only in internal functions right?

R/tbl_ancova.R
```diff
@@ -0,0 +1,301 @@
...
+#' theme_gtsummary_roche()
```
Melkiades: not needed

R/tbl_ancova.R
```diff
@@ -0,0 +1,301 @@
...
+  emm_summary <- summary(emm, calc = c(n = ".wgt.")) |>
+    dplyr::as_tibble() |>
+    dplyr::rename(
+      estimate = dplyr::any_of(c("emmean", "prob")),
+      n = dplyr::any_of("n")
+    )
+
+  # contrasts (treatment vs control) -------------------------------------------
+  contr_summary <- emmeans::contrast(emm, method = "trt.vs.ctrl", ref = ref_group, adjust = adjust) |>
+    summary(infer = TRUE, level = conf.level) |>
```
Melkiades, line 170: is this from stats?

R/tbl_ancova.R
```diff
@@ -0,0 +1,301 @@
...
+    dplyr::rename(
+      conf.low = dplyr::any_of(c("lower.CL", "asymp.LCL")),
+      conf.high = dplyr::any_of(c("upper.CL", "asymp.UCL"))
+    )
+
+  # parse contrast labels to extract the non-reference group name
+  contr_summary$trt_group <- sub(
+    paste0("\\s*-\\s*", .escape_regex(ref_group), "$"), "",
+    contr_summary$contrast
+  )
```
Melkiades, line 183: add a comment with an example value and its extracted value

R/tbl_ancova.R
```diff
@@ -0,0 +1,301 @@
...
+
+  # build ARD for tbl_ard_summary ----------------------------------------------
+  trt_levels <- levels(data[[by]])
```
Melkiades, line 186: can this not be a factor?

R/tbl_ancova.R
```diff
@@ -0,0 +1,301 @@
...
+  endpoint_label <- all.vars(formula)[1]
```
Melkiades: how does this work exactly?

R/tbl_ancova.R
```diff
@@ -0,0 +1,301 @@
...
+    conf.level = conf.level
+  )
+
+  # build gtsummary table ------------------------------------------------------
+  ci_pct <- paste0(round(conf.level * 100), "%")
+
+  tbl <- gtsummary::tbl_ard_summary(
+    cards = ard,
+    by = dplyr::all_of(by),
+    type = everything() ~ "continuous2",
```
Melkiades, line 206: remind me why it is better to select continuous2? and if this is aligned with crane theme

R/tbl_ancova.R
```diff
@@ -0,0 +1,301 @@
...
+                "p-value"
+              ) & grepl("NA", .x, fixed = TRUE),
+              "",
+              .x
+            )
+          )
+        )
+    )
+
+  # add denominator header Ns if provided
```
Melkiades, line 243: add an example in the docs about this

R/tbl_ancova.R
```diff
@@ -0,0 +1,301 @@
...
+      )
+      tbl <- tbl |>
+        gtsummary::modify_header(!!!header_labels)
+    }
+  }
+
+  # add class and attributes
+  class(tbl) <- c("tbl_ancova", class(tbl))
+  attr(tbl, "ref_group") <- ref_group
+  attr(tbl, "conf.level") <- conf.level
```
Melkiades: I think there is usually the by, right?

R/tbl_ancova.R
```diff
@@ -0,0 +1,301 @@
...
+
+  tbl
+}
+
+
+# Internal: build the ARD data frame for tbl_ard_summary -----------------------
+.build_ancova_ard <- function(emm_summary, contr_summary, by,
```
Melkiades: describe better how it works

Melkiades: as a comment

R/tbl_ancova.R
```diff
@@ -0,0 +1,301 @@
...
+
+
+# Internal: escape regex special characters ------------------------------------
+.escape_regex <- function(x) {
+  chars <- c(".", "|", "^", "$", "(", ")", "[", "]", "{", "}", "*", "+", "?", "\\")
+  for (ch in chars) {
+    x <- gsub(ch, paste0("\\", ch), x, fixed = TRUE)
+  }
+  x
+}
```
Melkiades: give me an example as a comment for special cases

R/tbl_ancova.R
```diff
@@ -0,0 +1,301 @@
...
+        .ancova_ard_row(endpoint_label, by, trt, "conf.low", NA_real_, "CI Lower Bound", "numeric"),
+        .ancova_ard_row(endpoint_label, by, trt, "conf.high", NA_real_, "CI Upper Bound", "numeric"),
+        .ancova_ard_row(endpoint_label, by, trt, "p.value", NA_real_, "p-value", "numeric")
+      ))
+    }
+  }
+
+  dplyr::bind_rows(ard_rows) |>
+    dplyr::mutate(
+      context = "continuous",
```
Melkiades, line 329: why?

Melkiades: check that is used before any processing

R/tbl_ancova.R
```diff
@@ -0,0 +1,301 @@
...
+      fmt_fun = purrr::map2(.data$stat_name, .data$stat, function(nm, val) {
```
Melkiades: my ignorance: why did you use fmt_fun? and map2?

tests/testthat/test-tbl_ancova.R
```diff
@@ -0,0 +1,114 @@
...
+    dplyr::filter(.data$label == "Difference in Adjusted Means") |>
+    dplyr::pull("stat_1")
+  expect_equal(ref_col, "")
+
+  expect_snapshot(as.data.frame(tbl))
+})
+
+
+test_that("tbl_ancova() works with denominator", {
+  expect_silent(
```
Melkiades, line 44: check the numbers too

tests/testthat/test-tbl_ancova.R
```diff
@@ -0,0 +1,114 @@
...
+  )
+
+  # CI label reflects 90%
+  labels <- tbl$table_body$label[tbl$table_body$row_type == "level"]
+  expect_true(any(grepl("90%", labels)))
+})
+
+
+test_that("tbl_ancova() errors on invalid ref_group", {
+  expect_error(
```
Melkiades: ALWAYS check errors with snapshots

tests/testthat/test-tbl_ancova.R
```diff
@@ -0,0 +1,114 @@
...
+    tbl <- tbl_ancova(
+      data = df_ancova,
+      formula = CHG ~ TRTA + BASE,
+      by = TRTA,
+      ref_group = "Placebo",
+      adjust = "dunnett"
+    )
+  )
+
+  expect_s3_class(tbl, "tbl_ancova")
```
Melkiades, line 201: should we check something more? like the label values? where should we track the method used?

R/tbl_ancova.R
```diff
@@ -45,12 +45,7 @@
 #'   When supplied, the column headers show `(N = <count>)` from this data
 #'   frame rather than from `data`.
 #'
-#' @return A `'gtsummary'` table of class `c("tbl_ancova", "gtsummary")`.
```
Melkiades: return should be here

R/tbl_ancova.R
```diff
@@ -60,6 +55,29 @@
...
+#'   tbl_ancova(
+#'     formula = CHG ~ TRTA + BASE,
+#'     by = TRTA,
+#'     ref_group = "Placebo",
+#'     adjust = "dunnett"
+#'   )
+#'
+#' @return A `'gtsummary'` table of class `c("tbl_ancova", "gtsummary")`.
+#' @name tbl_ancova
+#'
```
Melkiades: no space. See other parts of the code

R/tbl_ancova.R
```diff
@@ -291,7 +317,8 @@ tbl_ancova <- function(data,
 }
 
 
-# Internal: escape regex special characters ------------------------------------
+# Escapes regex special characters so group names can be used in sub() patterns.
+# e.g. "Arm (High)" -> "Arm \\(High\\)", "Dose 1.5mg" -> "Dose 1\\.5mg"
 .escape_regex <- function(x) {
   chars <- c(".", "|", "^", "$", "(", ")", "[", "]", "{", "}", "*", "+", "?", "\\")
```
Melkiades, line 367: very smart

R/tbl_ancova.R
```diff
@@ -209,14 +229,20 @@ tbl_ancova <- function(data,
 
   # add class and attributes
   class(tbl) <- c("tbl_ancova", class(tbl))
+  attr(tbl, "by") <- by
   attr(tbl, "ref_group") <- ref_group
   attr(tbl, "conf.level") <- conf.level
+  attr(tbl, "adjust") <- adjust
+  attr(tbl, "method") <- method
```
Melkiades: maybe also package? could you show also an example with this changed?

Melkiades: other function have this all as a single call ref (in gtsummary). could you check that?

R/tbl_ancova.R
Melkiades: At this point it's always a factor — line 103 does data[[by]] <- stats::relevel(factor(data[[by]]), ref = ref_group). So levels() is guaranteed to work.

R/tbl_ancova.R
Melkiades: No. It dispatches to emmeans:::summary.emmGrid(). infer = TRUE tells emmeans to compute CIs and p-values for the contrasts. level = conf.level sets the CI width. It's an emmeans-specific S3 method.

R/tbl_ancova.R
Melkiades: all.vars() extracts variable names from a formula. For CHG ~ TRTA + BASE it returns c("CHG", "TRTA", "BASE"). [1] takes the response variable (CHG), which becomes the row label in the table.

R/tbl_ancova.R
Melkiades: continuous2 allows multiple statistic rows per variable (n, adjusted mean, difference, CI, p-value — 5 rows stacked). Regular "continuous" renders a single summary line. Same pattern as tbl_baseline_chg and tbl_mmrm. Aligned with crane theme — tbl_roche_summary defaults to continuous2.

R/tbl_ancova.R
Melkiades: Required by cards::as_card(). It tells tbl_ard_summary() how to interpret the ARD rows. Without it, the rendering engine doesn't know the statistic type.

R/tbl_ancova.R
```diff
@@ -258,7 +284,7 @@ tbl_ancova <- function(data,
   dplyr::bind_rows(ard_rows) |>
     dplyr::mutate(
       context = "continuous",
-      fmt_fun = purrr::map2(.data$stat_name, .data$stat, function(nm, val) {
+      fmt_fun = purrr::map(.data$stat_name, function(nm) {
         switch(nm,
```
Melkiades: fmt_fun is the column cards::as_card() uses to store per-row formatting functions (n → integer, estimate → 2 decimals, p.value → 4 decimals). map2 was unnecessary since only stat_name drives the switch() — simplified to purrr::map().

R/tbl_ancova.R
```diff
@@ -242,7 +242,7 @@ tbl_ancova <- function(data,
 
 # Assembles an ARD (Analysis Results Dataset) from emmeans summaries.
 # For each treatment level, creates rows for: n, adjusted mean (from emm_summary),
-# and contrast stats — mean difference, CI bounds, p-value (from contr_summary).
+# and contrast stats <U+2014> mean difference, CI bounds, p-value (from contr_summary).
```
Melkiades: just write :

R/tbl_ancova.R
```diff
@@ -0,0 +1,330 @@
...
+
+  dplyr::bind_rows(ard_rows) |>
+    dplyr::mutate(
+      context = "continuous",
+      fmt_fun = map(.data$stat_name, function(nm) {
+        switch(nm,
+          "n" = function(x) sprintf("%.0f", x),
+          "estimate" = ,
+          "mean.diff" = ,
+          "conf.low" = ,
```
Melkiades, line 335: In R's switch(), empty assignments act as fall-throughs — they share the next non-empty result.

Melkiades:
Fixed a column-swap bug in `tbl_ancova()` when the reference group is not alphabetically first (e.g. `"B: Placebo"` vs `"A: Drug X"`).

**Root cause:** `tbl_ard_summary()` assigns `stat_1`, `stat_2`, etc. alphabetically by group level. The denominator header logic assumed factor level order (ref first via `relevel()`). When these orders diverged, headers were written to the wrong columns — values appeared under incorrect treatment arms.

**Fix (commit a86b83e):** The denominator header logic now reads the actual column-to-group mapping from `tbl$table_styling$header` instead of assuming positional order. Added a regression test with non-alphabetical arm names and an example showing how to control column order via factor levels on the input data.

Melkiades:
recheck

Melkiades:
Review compared with camis and user sas

### Statistical method validation

The approach is correct and matches SAS. I verified against the CAMIS ANCOVA comparison page:

- **LS means**: `lm()` + `emmeans::emmeans()` produces identical LS means to SAS `PROC GLM / LSMEANS` (CAMIS confirms: drugA=6.71, drugD=6.82, drugF=10.16 — all match).
- **Contrasts**: `emmeans::contrast(method = "trt.vs.ctrl")` is the right choice for treatment-vs-control (not pairwise). This matches the standard ANCOVA table layout.
- **`n` extraction**: `summary(emm, calc = c(n = ".wgt."))` correctly returns per-group observation counts.
- **Multiplicity**: `adjust = "none"` as default is a deliberate override of emmeans' default (which is Dunnett for `trt.vs.ctrl`). This is fine for the typical single-parameter ANCOVA table, but worth documenting — users expecting emmeans' default behavior might be surprised.

CAMIS summary: *"The R ANCOVA analysis is comparable to SAS. Comparison between SAS and R show identical results."*

### Issues

#### 1. Contrast label parsing breaks with hyphenated group names (bug)

`emmeans` wraps group names containing hyphens in parentheses: `"Drug-X"` becomes `"(Drug-X) - Placebo"` in the contrast label. The `sub()` extraction produces `"(Drug-X)"`, which doesn't match the factor level `"Drug-X"`. Result: `contr_row` has 0 rows, contrast stats silently disappear.

```r
# Reproducer:
df <- data.frame(y = rnorm(30), arm = factor(rep(c("Drug-X", "Drug-Y", "Placebo"), each = 10)))
# ... tbl_ancova will silently drop the contrast rows for Drug-X and Drug-Y

PR #234, opened 2026-05-11
feat: emit zero-rows for unobserved factor levels in tbl_hierarchical_rate_and_count()

### LINKED ISSUE #233: tbl_hierarchical_rate_and_count(): emit zero-rows for unobserved factor levels
## Problem

When a hierarchical variable (e.g., `basket_name`) is a factor with levels that have no observations in the data, `tbl_hierarchical_rate_and_count()` silently drops those levels. This forces downstream templates (e.g., AET02_SMQ in tlg.templates) to build elaborate workarounds: creating dummy data, running the function on it, then zeroing out the ARD — or manually injecting zero-rows via `modify_table_body()`.

## Proposed solution

Respect factor levels in the first hierarchical variable. If `variables[1]` is a factor, emit zero-rows for unobserved levels with `"0"` in all stat columns. The zero-rows should include:
- A label row for the unobserved level (the basket/SOC header)
- A rate summary row (`label_rate`) with `"0"`
- A count summary row (`label_count`) with `"0"`

This mirrors the structure that observed levels produce, just with zero counts.

## Root cause

`gtsummary::tbl_hierarchical()` passes hierarchical variables as `strata` to `cards::ard_tabulate()`, which only retains observed combinations. The `by` parameter retains all factor levels, but hierarchical variables are never passed as `by`.

## Expected usage

```r
df_adae_long |>
  dplyr::mutate(basket_name = factor(basket_name, levels = all_basket_levels)) |>
  tbl_hierarchical_rate_and_count(
    variables = c(basket_name, AEDECOD),
    denominator = adsl,
    by = TRT01A
  )
# -> zero-rows emitted for baskets in `all_basket_levels` with no events
```

## Impact

This would eliminate ~50 lines of workaround code in the AET02_SMQ template and make the pattern reusable for any template with potentially empty hierarchical groups.

Melkiades:
* `tbl_hierarchical_rate_and_count()` now emits zero-rows for unobserved factor levels in the first hierarchical variable. (#233, @Melkiades)

When `variables[1]` is a factor, any levels not present in the data get a header row (NA stats), a rate summary row ("0"), and a count summary row ("0") appended to the table. This eliminates the need for dummy-data workarounds in templates where some hierarchical groups may have no matching events (e.g., empty SMQ baskets in AET02_SMQ).

The implementation is a post-processing step (`.inject_unobserved_levels()`) that runs after the table is fully built. Non-factor variables are ignored — existing behavior is unchanged.

**Reference GitHub issue associated with pull request.** closes #233

- [ ] Request a reviewer

### DIFF
--- a/R/tbl_hierarchical_rate_and_count.R
+++ b/R/tbl_hierarchical_rate_and_count.R
@@ -340,13 +340,17 @@ add_overall.tbl_hierarchical_rate_and_count <- function(x,
 
   # only act when the column is a factor with defined levels
 
-  if (!is.factor(col_vals)) return(tbl)
+  if (!is.factor(col_vals)) {
+    return(tbl)
+  }
 
   observed <- unique(as.character(col_vals))
   all_levels <- levels(col_vals)
   missing_levels <- setdiff(all_levels, observed)
 
-  if (length(missing_levels) == 0L) return(tbl)
+  if (length(missing_levels) == 0L) {
+    return(tbl)
+  }
 
   cli::cli_inform(
     c("i" = "Adding zero-rows for {length(missing_levels)} unobserved level{?s} in {.val {top_var}}.")
--- a/R/tbl_hierarchical_rate_and_count.R
+++ b/R/tbl_hierarchical_rate_and_count.R
@@ -12,6 +12,21 @@
 #' `gtsummary::tbl_hierarchical_count()` directly, followed by a call to
 #' `gtsummary::filter_hierarchical()`.
 #'
+#' @details
+#' ## Factor levels and zero-rows
+#'
+#' When the first variable in `variables` is a factor, the function respects
+#' its levels and emits zero-rows for any levels not observed in the data.
+#' Each unobserved level gets a header row (with `NA` stats), a rate summary
+#' row (`"0"`), and a count summary row (`"0"`).
+#'
+#' This is useful when the set of expected categories is predefined but some
+#' may not be present in the data. To exclude categories with no observations,
+#' drop unused levels before calling the function (e.g., `droplevels()`).
+#'
+#' Ensures consistent output structure for empty datasets, removing the
+#' need for manual workarounds in reporting templates.
+#'
 #' @inheritParams gtsummary::tbl_hierarchical
 #' @inheritParams gtsummary::add_overall.tbl_hierarchical
 #' @param variables ([`tidy-select`][dplyr::dplyr_tidy_select])\cr
@@ -51,6 +66,18 @@
 #'     by = TRTA
 #'   ) |>
 #'   add_overall(last = TRUE)
+#'
+#' # Example 2: factor with unobserved levels ----------------------------------
+#' # Adding an unobserved SOC level produces zero-rows automatically
+#' cards::ADAE[c(1, 2, 3, 8, 16), ] |>
+#'   dplyr::mutate(
+#'     AEBODSYS = factor(AEBODSYS, levels = c(unique(AEBODSYS), "UNOBSERVED SOC"))
+#'   ) |>
+#'   tbl_hierarchical_rate_and_count(
+#'     variables = c(AEBODSYS, AEDECOD),
+#'     denominator = cards::ADSL,
+#'     by = TRTA
+#'   )
 NULL
 
 #' @rdname tbl_hierarchical_rate_and_count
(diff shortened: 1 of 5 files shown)


R/tbl_hierarchical_rate_and_count.R
```diff
@@ -6,6 +6,11 @@
 #' The function produces additional summary rows for the higher level nesting
 #' variables providing both rates and counts.
 #'
+#' When the first variable in `variables` is a factor, zero-rows are emitted
+#' for any unobserved levels. This allows templates to define the expected
+#' categories upfront (e.g., SMQ baskets) and get complete output without
+#' post-processing.
```
Melkiades: Describe better, that this happens when there are nor data rows and this is a special case. this stuff belongs to details. Add a small example too

R/tbl_hierarchical_rate_and_count.R
```diff
@@ -239,6 +264,18 @@ tbl_hierarchical_rate_and_count <- function(data,
...
+  # When variables[1] is a factor, emit zero-rows for levels with no events.
+  # gtsummary::tbl_hierarchical() drops unobserved strata levels, so we
+  # re-inject them here to keep the output complete.
+  tbl_final <- .inject_unobserved_levels(
+    tbl_final,
+    data = data,
+    variables = variables,
+    label_rate = label_rate,
+    label_count = label_count
+  )
```
Melkiades, line 299: do you think this should always happen? I think it is better to keep it as it is easy to drop levels in pre-proc if needed. Question: should that be the behavior of tbl_hierarchical?

R/tbl_hierarchical_rate_and_count.R
Melkiades: Updated: moved to `@details` with a section explaining both cases (unobserved levels and 0-row data), added a note about `droplevels()` for users who don't want this, and added Example 2 showing the factor-level behavior.

R/tbl_hierarchical_rate_and_count.R
Melkiades: Yes, I think always-on is the right default — it follows the R convention that factors carry their levels. If a user passes a factor, they're declaring intent about the expected categories. If they don't want zero-rows, `droplevels()` in pre-proc is a one-liner.

Re: `tbl_hierarchical` — ideally yes, this belongs upstream. The root cause is that `cards::ard_hierarchical()` passes hierarchical variables as `strata` (observed-only) instead of `by` (all factor levels). But that's a gtsummary/cards change with broader implications. This crane-level post-processing is a pragmatic solution that doesn't require upstream changes.

R/tbl_hierarchical_rate_and_count.R
```diff
@@ -12,6 +12,23 @@
...
+#'
+#' When the first variable in `variables` is a factor, the function respects
+#' its levels and emits zero-rows for any levels not observed in the data.
+#' Each unobserved level gets a header row (with `NA` stats), a rate summary
+#' row (`"0"`), and a count summary row (`"0"`).
+#'
+#' This is useful when the set of expected categories is known upfront
+#' (e.g., SMQ baskets from an AAG dataset) but some categories may have no
+#' matching events. Users who do not want this behavior can drop unused
+#' levels before calling the function (e.g., `droplevels()`).
```
Melkiades: be more generic.

R/tbl_hierarchical_rate_and_count.R
```diff
@@ -12,6 +12,23 @@
...
+#'
+#' When the data has zero rows and the first variable is a factor, a scaffold
+#' table is built with correct column headers and zero-rows for all levels.
+#' This avoids the need for dummy-data workarounds in templates.
```
Melkiades: too many internal details maybe? be more direct

R/tbl_hierarchical_rate_and_count.R
jszczypinski:
```suggestion
#' This is useful when the set of expected categories is predefined but some 
#' may not be present in the data. To exclude categories with no observations, 
#' drop unused levels before calling the function (e.g., `droplevels()`).
```

R/tbl_hierarchical_rate_and_count.R
jszczypinski:
```suggestion
#' Ensures consistent output structure for empty datasets, removing the 
#' need for manual workarounds in reporting templates.
```

R/tbl_hierarchical_rate_and_count.R
```diff
@@ -100,6 +129,26 @@ tbl_hierarchical_rate_and_count <- function(data,
   # saving function inputs
   tbl_hierarchical_rate_and_count_inputs <- as.list(environment())
 
+  # handle 0-row data with factor levels ---------------------------------------
+  # gtsummary::tbl_hierarchical() cannot process empty data. When the first
+  # variable is a factor we build a scaffold table with zero-rows for all levels.
+  if (nrow(data) == 0L && is.factor(data[[variables[1]]])) {
```
jszczypinski, line 133: I wonder, whether we should make mandatory for variables to be a factor? Or to convert it to factor within the function and inform the user?

Melkiades: Good point!

R/tbl_hierarchical_rate_and_count.R
Melkiades: Applied your suggestion verbatim.

R/tbl_hierarchical_rate_and_count.R
Melkiades: Applied your suggestion verbatim.

R/tbl_hierarchical_rate_and_count.R
Melkiades: Implemented: character columns are now auto-converted to factor with a `cli::cli_inform()` nudging the user to set levels explicitly. Label attributes are preserved during conversion. The 0-row scaffold still only triggers for factors (since character-to-factor with 0 rows produces no levels).

Melkiades: Reverted the auto-conversion. Character columns are left as-is — no casting, no message. The `cli::cli_inform()` now fires only inside `.inject_unobserved_levels()` when unobserved factor levels are actually found and zero-rows are being added.

Melkiades: yes because that is the only case where you should produce that table

jszczypinski (approved):
LGTM :)

PR #244, opened 2026-05-18
191 193 coxph pvalue ci method

jszczypinski:
**What changes are proposed in this pull request?**
* `get_cox_pairwise_df` now allows for different methods to compute p value (#191) and to handle ties (#193)

Provide more detail here as needed.

Closes #191 
Closes #193 

`get_cox_pairwise_df` now allow to choose a method used to handle ties, with default to `exact`
Also, user can choose a method to estimate p value - all methods used in SAS LIFETEST Proc are available.
The latter required adding new dependency - {coin} as most of the tests available in SAS were not available in {survival} package.

--------------------------------------------------------------------------------

Pre-review Checklist (if item does not apply, mark is as complete)
- [ ] **All** GitHub Action workflows pass with a :white_check_mark:
- [ ] PR branch has pulled the most recent updates from master branch: `usethis::pr_merge_main()`
- [ ] If a bug was fixed, a unit test was added.
- [ ] Code coverage is suitable for any new functions/features (generally, 100% coverage for new code): `devtools::test_coverage()`
- [ ] Request a reviewer

Reviewer Checklist (if item does not apply, mark is as complete)

- [ ] If a bug was fixed, a unit test was added.
- [ ] Run `pkgdown::build_site()`. Check the R console for errors, and review the rendered website.
- [ ] Code coverage is suitable for any new functions/features: `devtools::test_coverage()`

When the branch is ready to be merged:
- [ ] Update `NEWS.md` with the changes from this pull request under the heading "`# cards (development version)`". If there is an issue associated with the pull request, reference it in parentheses at the end update (see `NEWS.md` for examples).
- [ ] **All** GitHub Action workflows pass with a :white_check_mark:
- [ ] Approve Pull Request
- [ ] Merge the PR. Please use "Squash and merge" or "Rebase and merge".

PR #245, opened 2026-05-19
207 tbl coxph work with get cox pairwise df

### LINKED ISSUE #207: Make tbl_coxph() to work with get_cox_pairwise_df()
This is a proposal. It would render more explicit some modifications before tbl_coxph. For examples if we want to modify <0.001 to <0.01 or apply other methods from survival.

jszczypinski:
* `tbl_coxph()` now accepts a data.frame, created with `get_cox_pairwise_df()` as input.

closes #207 
This PR should be merged first https://github.com/insightsengineering/crane/pull/244

- [ ] Request a reviewer

- [ ] Approve Pull Request


R/get_cox_pairwise_df.R
```diff
@@ -131,12 +183,63 @@ get_cox_pairwise_df <- function(model_formula, data, arm, ref_group = NULL) {
...
+    rlang::check_installed(
+      "coin",
+      reason = paste("to run log-rank tests using", test_type, "method")
+    )
+
+    test_result <- coin::logrank_test(
+      formula = formula,
+      data = data,
+      type = test_type
+    )
```
Melkiades, line 238: Pay attenion that `coin::logrank_test()` requires the arm variable to be a factor with exactly 2 levels in the subset

In `get_cox_pairwise_df()`, `comp_df` is filtered to 2 arms but the factor levels are not dropped:
```r
comp_df <- data[as.character(data[[arm]]) %in% subset_arm, ]
```
`comp_df[[arm]]` still carries all original factor levels (e.g., 3 levels for a 3-arm study). `coin::logrank_test()` may behave unexpectedly with unused factor levels. Add `droplevels()`:
```r
comp_df <- data[as.character(data[[arm]]) %in% subset_arm, ]
comp_df[[arm]] <- droplevels(comp_df[[arm]])
```

Melkiades: I would add a warning if that happens

R/get_cox_pairwise_df.R
```diff
@@ -131,12 +183,63 @@ get_cox_pairwise_df <- function(model_formula, data, arm, ref_group = NULL) {
       pval = log_rank_pvalue
     )
     rownames(current_row) <- current_arm
```
Melkiades, line 182: This is unrelated to your additions but it could be helpful to tackle in another PR:

`conf_int_row` lookup is fragile

```r
conf_int_row <- paste0(arm, current_arm)
```
This constructs `"armB"` to index into `coxph_ans$conf.int`. But `arm` is the column name string (e.g., `"arm"`), not the variable name in the model. If the user's column is named `"TRT01A"`, the row name would be `"TRT01AXanomeline High Dose"`. This works, but only because `coxph` names coefficients as `<varname><level>`. If the formula has covariates (e.g., `Surv(time, status) ~ arm + age`), the `conf.int` matrix has multiple rows and the indexing still works because it's by row name. This is fine but worth a comment.

R/get_cox_pairwise_df.R
```diff
@@ -97,7 +147,7 @@ get_cox_pairwise_df <- function(model_formula, data, arm, ref_group = NULL) {
   }
   comp_group <- setdiff(levels(data[[arm]]), ref_group)
 
-  ret <- c()
+  res <- c()
```
Melkiades: I would make it a list() and then `do.call(rbind, ...)`

R/get_cox_pairwise_df.R
```diff
@@ -131,12 +183,63 @@ get_cox_pairwise_df <- function(model_formula, data, arm, ref_group = NULL) {
...
-    ret <- rbind(ret, current_row)
+    res <- rbind(res, current_row)
   }
-  names(ret) <- c(
+  names(res) <- c(
     "HR",
     "95% CI",
-    "p-value (log-rank)"
+    paste0("p-value (", names(log_rank_pvalue), ")")
   )
```
Melkiades, line 192: This is seems outside the loop. It will pick only last value

Melkiades: this works because all iterations use the same `test`, so the name is always the same. But it's fragile — if the loop body ever changes, this could silently break. Move the column naming inside the loop or compute the p-value column name once before the loop:
```r
pval_col_name <- paste0("p-value (", 
  if (test == "log-rank") test else tools::toTitleCase(test), ")")
```

R/get_cox_pairwise_df.R
```diff
@@ -131,12 +183,63 @@ get_cox_pairwise_df <- function(model_formula, data, arm, ref_group = NULL) {
...
+    )
+
+    test_result <- coin::logrank_test(
+      formula = formula,
+      data = data,
+      type = test_type
+    )
+
+    p_value <- coin::pvalue(test_result)
+  } else if (test_type == "lr") {
```
Melkiades: always true when reached. Simplify in else

tests/testthat/test-tbl_coxph.R
```diff
@@ -16,105 +16,122 @@ surv_data_3arm <- survival::lung |>
...
-test_that("tbl_coxph() works with 2 arms and uses default ref_group", {
-  form <- survival::Surv(time, status) ~ arm
-
-  res <- tbl_coxph(
-    data = surv_data_2arm,
-    model_formula = form,
-    arm = "arm"
-    # Omitting ref_group to test the default `levels(data[[arm]])[1]` branch
-  )
+test_that("tbl_coxph() works with single comparison (2 arms)", {
```
Melkiades, line 53: Missing `skip_if_not_installed("coin")` in `test-tbl_coxph.R`?

R/get_cox_pairwise_df.R
jszczypinski: Moved it outside of the loop

R/get_cox_pairwise_df.R
jszczypinski: Done :)

R/get_cox_pairwise_df.R
jszczypinski: I made another issue tracking that #247

R/get_cox_pairwise_df.R
jszczypinski: I added droplevels and also added a warning

Melkiades (changes requested):
wait for the merge from main, I will merge it in smoothly afterwards ;) Thanks anyway :) it looks good

NEWS.md
```diff
@@ -30,7 +30,9 @@
 * Added `adjust_stat_columns_wrap()` which keep the presentation of statistics in one row (#219)
 
 * `get_cox_pairwise_df` now allows for different methods to compute p-value (#191) and to handle ties (#193)
- 
+
+* `tbl_coxph()` now accepts a data.frame, created with `get_cox_pairwise_df()` as input
```
Melkiades:
```suggestion
* `tbl_coxph()` now accepts a data.frame, created with `get_cox_pairwise_df()` as input (#207)
```

Someone got into a large context ahah

R/tbl_coxph.R
```diff
@@ -1,33 +1,22 @@
...
+#' @param pairwise_df (`data.frame`)\cr
+#'   The results `data.frame` output generated by `get_cox_pairwise_df()`.
+#'   Must contain rownames (comparison arms) and at least one statistic column:
+#'   `"HR"`, `"95% CI"`, or `"p-value (...)"`.
 #'
 #' @return A `gtsummary` object summarizing the pairwise Cox PH results.
 #'
+#' @seealso `get_cox_pairwise_df()`.
+#'
 #' @examplesIf requireNamespace("survival", quietly = TRUE)
```
Melkiades: maybe also coin? if you used it only in another part add another examplesIf

NEWS.md
```diff
@@ -30,7 +30,9 @@
 * Added `adjust_stat_columns_wrap()` which keep the presentation of statistics in one row (#219)
 
 * `get_cox_pairwise_df` now allows for different methods to compute p-value (#191) and to handle ties (#193)
```
Melkiades: From other PR: the return still says pval: The log-rank p-value — should reflect the dynamic column name p-value (<test>) and the 7 supported test types.

R/tbl_coxph.R
```diff
@@ -159,51 +129,72 @@ tbl_coxph <- function(data, model_formula, arm, ref_group = NULL) {
...
+  if (has_ci) {
+    vars_include <- c(vars_include, "ci_formatted")
+    label_list <- c(label_list, list(ci_formatted ~ "95% CI"))
+  }
+
+  # Construct the base table
+  res <- data_subset |>
     cards::ard_mvsummary(
-      variables = c("pval_formatted", "hr_formatted", "ci_formatted"),
+      variables = dplyr::all_of(vars_include),
```
Melkiades, line 175: So this comes probably from my work but I think we could highlight that using ard_mvsummary() with my_stat = \(x, ...) x[1] to extract a pre-formatted string into an ARD, then rendering via tbl_ard_summary, is non-obvious. A one-line comment explaining why this roundabout approach is needed (to get a proper gtsummary object with ARD backing) would help future readers.

R/tbl_coxph.R
```diff
@@ -36,112 +25,93 @@
...
+    df_tidy$ci_formatted <- pairwise_df[["95% CI"]]
+  }
+  if (has_pval) {
+    df_tidy$p_num <- pairwise_df[[pval_label]]
+    df_tidy <- df_tidy |>
+      dplyr::mutate(
+        pval_formatted = dplyr::case_when(
+          is.na(.data$p_num) ~ NA_character_,
+          .data$p_num < 0.0001 ~ "<0.0001",
+          TRUE ~ sprintf("%.4f", .data$p_num)
```
Melkiades: The p-value formatting (<0.0001 / sprintf("%.4f", ...)) is hardcoded. If users need different precision, they'd have to modify the data frame before passing it. Acceptable for now.

Maybe a new issue to discuss?

Melkiades: I do not think we should add parameters for this as you know so I would consider in the future to make an example on how to correct the above...

Melkiades: actually thinking about it - this is in pre-proc, could we add it as a formatting function? so it is very easy to post-process eventually

R/get_cox_pairwise_df.R
```diff
@@ -31,32 +31,26 @@
...
-#'   (excluding the reference group). For each iteration, it filters the data
-#'   to include only the current comparison arm and the reference arm, and then:
+#' @details The function iterates through each non-reference arm, subsets the
+#'   data to the current arm and the reference arm, and then:
 #'   \itemize{
-#'     \item Fits a Cox model using `survival::coxph`.
-#'     \item Performs a log-rank test using `survival::survdiff`.
+#'     \item Fits a Cox model using [survival::coxph()].
+#'     \item Computes a p-value via [.estimate_p_value()], which dispatches
+#'       to [coin::logrank_test()] for weighted log-rank variants or to
```
Melkiades: is this going to be an issue if in suggests?

R/tbl_coxph.R
```diff
@@ -198,3 +191,21 @@ tbl_coxph <- function(pairwise_df) {
...
+
+#' Format p-values for display
+#'
+#' Values below 0.0001 are shown as `"<0.0001"`, others are formatted to
+#' four decimal places. Override this function or post-process the
+#' `pval_formatted` column in `df_tidy` if different precision is needed.
+#'
+#' @param x (`numeric`)\cr vector of p-values.
+#'
+#' @returns A character vector of formatted p-values.
```
Melkiades, line 215: question: is this the best way? is it intuitive for users considering how it is commonly done with gtsummary and in crane?

Melkiades: once decided for a way we should document it properly, like a small vignette that also contains the logic of tbl_coxph (for example) -> statistics - tabulation - post-proc/esthetics. When would it best for users to set the digits for example? Is this modularity helping?

Melkiades:
Pushed `d1c029f5` addressing all open review comments. Summary:

**Docs**
- Fixed stale `@details`/`@seealso`/`@return` in `get_cox_pairwise_df()` — removed `survdiff` references, updated return column names to reflect dynamic `p-value (<test>)` naming.
- Added `@param`/`@returns` to `.estimate_p_value()`.
- Updated `tbl_coxph()` `@return` to mention the `tbl_coxph` class.
- Cleaned up NEWS entries.

**Code**
- Extracted p-value formatting into `.format_pvalue()` helper.
- Added comment explaining the `ard_mvsummary()` roundabout in `.get_single_comp_table()`.
- Added `coin` to `@examplesIf` guard in `tbl_coxph()`.

---

**Re: `coin` hyperlink in `@details`/`@seealso`** — you're right, if `coin` is not installed the `[coin::logrank_test()]` roxygen link will be broken (dead link in rendered docs, R CMD check NOTE). I'll replace it with plain backtick references (`` `coin::logrank_test()` ``) in the next commit — no hyperlink, no broken link.

**Re: `.format_pvalue()` approach** — keeping Option A (internal helper). Adding an example to `tbl_coxph()` docs showing how to customize p-value precision via `modify_fmt_fun()` after the fact — this is the gtsummary-native way and gives users full control without adding API surface.

**Re: vignette** — Open a separate issue + PR for a modularity vignette covering the compute → tabulate → post-process pipeline (`get_cox_pairwise_df()` → `tbl_coxph()` → `modify_*()`), including when to set digits at each layer.

Melkiades (approved):

PR #254, opened 2026-05-25
Adding discrete ties method to get_cox_pairwise_df()

### LINKED ISSUE #193: get_cox_pairwise_df - No options for handling ties and incorrect default method
### Feature description

The GDSR indicates that, when computing the hazard ratio, we should use the discrete option to handle ties. However, the `get_cox_pairwise_df` function uses the effron method as default. My understanding is that we would want to set the default to the "exact" method, which seems to correspond to the "discrete" method outlined in the GDSR for SAS based studies. Regardless, the GDSR also indicates that we should allow the user to update this as required so more flexibility is required here to allow for different options. 

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [x] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [x] I agree to follow this project's Security Policy.

jszczypinski:
"There should be an option to specify the handling of ties with /TIES=DISCRETE as default (SAS options are: EXACT, BRESLOW, EFRON, DISCRETE)."

And the review from Rees, in which he stated that we have to change ties to exact. However after more digging I believe we have to add more code and add the option to include discrete ties as it is in SAS. This should be the default based on GDSR and it needs some more development in {crane}. The discrete method is not available in {survival} but it can be derived using stats::glm().

Closes #193

- [ ] Request a reviewer

- [ ] Approve Pull Request


R/get_cox_pairwise_df.R
```diff
@@ -250,10 +246,80 @@ get_cox_pairwise_df <- function(
...
+#' into person-time format, fits a logistic regression model, and calculates
+#' Wald confidence intervals.
+#'
+#' @param formula (`formula`)\cr
+#'   The original survival formula.
+#' @param data (`data.frame`)\cr
+#'   The subsetted data containing exactly two arms.
+#'
+#' @returns A list mimicking `summary(coxph)` containing a `conf.int` data.frame
+#' @keywords internal
```
llrs-roche: Internal documentation is available on the manual. Either make sure that `devtools::check_doc_fields()` pass or make it `@noRd` .

R/get_cox_pairwise_df.R
```diff
@@ -250,10 +246,80 @@ get_cox_pairwise_df <- function(
...
+.get_discrete_cox <- function(formula, data) {
+  # Sanitize to allow smooth parsing of the Surv formulation
+  lhs_str <- paste(deparse(formula[[2]]), collapse = " ")
+  rhs_str <- paste(deparse(formula[[3]]), collapse = " ")
+
+  clean_lhs_str <- sub(".*Surv\\(", "Surv(", lhs_str)
+  clean_formula <- stats::as.formula(paste(clean_lhs_str, "~", rhs_str))
```
llrs-roche: I think the standard way to modify formulas is to use `update.formula`, it might be more flexible.
Or using something along these lines:

```r
terms <- terms(formula) # Notice the arguments keep.order and simplify
terms_characters <- as.character(terms)
```

R/get_cox_pairwise_df.R
```diff
@@ -250,10 +246,80 @@ get_cox_pairwise_df <- function(
...
+#'   The subsetted data containing exactly two arms.
+#'
+#' @returns A list mimicking `summary(coxph)` containing a `conf.int` data.frame
+#' @keywords internal
+.get_discrete_cox <- function(formula, data) {
+  # Sanitize to allow smooth parsing of the Surv formulation
+  lhs_str <- paste(deparse(formula[[2]]), collapse = " ")
+  rhs_str <- paste(deparse(formula[[3]]), collapse = " ")
+
+  clean_lhs_str <- sub(".*Surv\\(", "Surv(", lhs_str)
```
llrs-roche: What happens when something else is used on the formula instead of {survival}?
I've also have read some discussions about the implications of using the package prefix and without it on formulas: Mainly this [recent thread](https://stat.ethz.ch/pipermail/r-devel/2025-April/083989.html).
I would advocate to remove them (perhaps with a warning) regardless of which package they come from.

R/get_cox_pairwise_df.R
llrs-roche: See comment about formulas, but this doesn't seem to be a good approach to set formulas: Functions inside formulas shouldn't be called with the namespace of the package which turns them into special objects.

```suggestion
#' formula <- Surv(time, status) ~ arm
#' formula <- `survival::Surv`(time, status) ~ arm
```

R/get_cox_pairwise_df.R
```diff
@@ -250,10 +246,80 @@ get_cox_pairwise_df <- function(
...
+  clean_formula <- stats::as.formula(paste(clean_lhs_str, "~", rhs_str))
+
+  mf <- stats::model.frame(clean_formula, data = data)
+  surv_resp <- stats::model.response(mf)
+
+  # Extract the actual status variable name directly from the formula.
+  # Surv() forces its internal matrix column names to "time" and "status",
+  # so we must bypass it and read the original variable name (e.g., 'is_event').
+  lhs_vars <- all.vars(formula[[2]])
+  status_var <- lhs_vars[length(lhs_vars)]
```
llrs-roche, line 297: I think this assumes that the formula is something on this order `time+var ~arm`. Or at least I understand it that way.

R/get_cox_pairwise_df.R
```diff
@@ -250,10 +246,80 @@ get_cox_pairwise_df <- function(
...
+  data_long <- survival::survSplit(
+    formula = clean_formula,
+    data = data,
+    cut = event_times,
+    episode = ".time_int"
+  )
+
+  glm_formula <- stats::as.formula(
+    paste(status_var, "~", rhs_str, "+ as.factor(.time_int)")
+  )
```
llrs-roche: I know that status might not always be the name of the status_var but check something like this and keep in mind the function calls on formulas:

```suggestion
  glm_formula <- update(clean_formula, status ~ . + `as.factor`(.time_int))
```

tests/testthat/test-get_cox_pairwise_df.R
```diff
@@ -112,15 +113,17 @@ test_that(paste0(
...
+  ties_methods <- c("discrete", "exact", "efron", "breslow")
 
   for (t_method in ties_methods) {
     expect_no_error(
-      res <- get_cox_pairwise_df(
-        model_formula = survival::Surv(time, status) ~ arm,
-        data = surv_data_2arm,
-        arm = "arm",
-        ties = t_method
+      suppressWarnings(
```
llrs-roche, line 121: Which warnings are being generated now that need to be supresed? This blank supresing of any warnings could lead to missing problems.

tests/testthat/test-get_cox_pairwise_df.R
```diff
@@ -112,15 +113,17 @@ test_that(paste0(
...
+        res <- get_cox_pairwise_df(
+          model_formula = survival::Surv(time, status) ~ arm,
```
llrs-roche: I would test with the formula being a bit different from the examples if possible

R/get_cox_pairwise_df.R
```diff
@@ -250,10 +246,80 @@ get_cox_pairwise_df <- function(
     fit_null <- survival::survreg(null_formula, data = data, dist = "exponential")
 
     # calculate the difference and estimate the statistics
-    lrt_stat <- 2 * (fit_cov$loglik[2] - fit_null$loglik[1])
+    lrt_stat <- 2 * (fit_cov$loglik[2] - fit_null$loglik[2])
```
llrs-roche, line 264: This could have an important impact on the outputs.  Why is the order changed?

R/get_cox_pairwise_df.R
```diff
@@ -250,10 +246,80 @@ get_cox_pairwise_df <- function(
...
+  )
+
+  glm_fit <- stats::glm(
+    glm_formula,
+    data = data_long,
+    family = stats::binomial(link = "logit")
+  )
+
+  coefs <- stats::coef(glm_fit)
+  wald_ci_log <- stats::confint.default(glm_fit)
```
llrs-roche, line 328: Do we want the default CI or the method specific one? 

```suggestion
  wald_ci_log <- confint(glm_fit)
```

R/get_cox_pairwise_df.R
```diff
@@ -250,10 +246,80 @@ get_cox_pairwise_df <- function(
...
+
+  lower_log_ci <- wald_ci_log[, "2.5 %"]
+  upper_log_ci <- wald_ci_log[, "97.5 %"]
+  hr <- exp(coefs)
+
+  conf_int_df <- data.frame(
+    "exp(coef)"  = hr,
+    "exp(-coef)" = 1 / hr,
+    "lower .95"  = exp(lower_log_ci),
+    "upper .95"  = exp(upper_log_ci),
```
llrs-roche, line 338: I am not sure of the mathematical transformation but this confidence interval creates one between 0 and Inf  while the exp(coef) is 8e-5.

R/get_cox_pairwise_df.R
```diff
@@ -222,15 +227,6 @@ get_cox_pairwise_df <- function(
       reason = paste("to run log-rank tests using", test_type, "method")
     )
 
-    if (length(levels(data[[arm]])) != 2) {
-      cli::cli_warn(
```
llrs-roche, line 226: Without this warning there might be erroneous output without notice. Is there a way to raise it when the pairwise test is used?

llrs-roche (commented):
The PR modifies the statistical approach of the function. 
It is back-compatible but the default is changes so result will break as the new methods will be used.
Many of my comments are focused on the formula manipulation which I am not sure are currently good enough.

R/get_cox_pairwise_df.R
jszczypinski: This was a bug that was not spotted in the previous PR. It won't have impact on outputs since likelihood ratio test was added last week and wasn't used before.

R/get_cox_pairwise_df.R
jszczypinski: It's impossible to trigger this warning since `comp_df[[arm]] <- droplevels(comp_df[[arm]])` in the main function forces exactly 2 levels.

tests/testthat/test-get_cox_pairwise_df.R
jszczypinski: Agree. I will change it.

R/get_cox_pairwise_df.R
jszczypinski: Thanks Lluis that was a great catch. I have updated the examples and also added a test to check if there is a namespace in the formula supplied by the user

R/get_cox_pairwise_df.R
jszczypinski: See comment below

R/get_cox_pairwise_df.R
jszczypinski: This part was removed

R/get_cox_pairwise_df.R
jszczypinski: We want Wald CIs as per GDSR programming notes. That's why I used default

R/get_cox_pairwise_df.R
jszczypinski: I believe this is correct. 
From Gem:
If exp(coef) = 0.00008, then the raw log coefficient is roughly -9.43.
If the standard error (SE) for this specific coefficient is enormous—say, 5000—your 95% Wald interval on the log scale would be approximately [-9809, 9790].
When you exponentiate those bounds in R:

exp(-9809) evaluates to 0

exp(9790) exceeds R's maximum floating-point limit and evaluates to Inf.

R/get_cox_pairwise_df.R
jszczypinski: I did something similar to your suggestion

R/get_cox_pairwise_df.R
jszczypinski: Done :)

R/get_cox_pairwise_df.R
```diff
@@ -55,17 +56,21 @@
 #' @examples
 #' # Example data setup (assuming 'time' is event time, 'status'
 #' # is event indicator (1=event), and 'arm' is the treatment group)
-#' library(dplyr) # For better data handling
+#' # for data handling
+#' library(dplyr)
+#' # to not specify formula using `survival::Surv` - it is a bad practice and
+#' # can lead to unexpected results
+#' library(survival)
```
Melkiades, line 61: is it still in suggests? I would add examplesif if so

R/get_cox_pairwise_df.R
```diff
@@ -250,10 +261,81 @@ get_cox_pairwise_df <- function(
...
+#' into person-time format, fits a logistic regression model, and calculates
+#' Wald confidence intervals.
+#'
+#' @param formula (`formula`)\cr
+#'   The original survival formula.
+#' @param data (`data.frame`)\cr
+#'   The subsetted data containing exactly two arms.
+#'
+#' @returns A list mimicking `summary(coxph)` containing a `conf.int` data.frame
+#' @noRd
```
Melkiades, line 285: noob question: is this showing if requested? I am wondering if we should have `@keywords internal` as always instead... ofc it is not like we use doc calls for internal functions anyway... (probably this is a better standard)

R/get_cox_pairwise_df.R
Melkiades: maybe I would add a comment about this

R/get_cox_pairwise_df.R
llrs-roche: Ah, I see. I hope it doesn't affect the snapshots elsewhere.

R/get_cox_pairwise_df.R
llrs-roche: It is in imports (on CRAN, and I think on this version too)

R/get_cox_pairwise_df.R
jszczypinski: We can revert to `@keywords internal` and add examples and so on. When there is `@noRd` tag the docs can still be requested using `?function_name`.

R/get_cox_pairwise_df.R
jszczypinski: Done :)

R/get_cox_pairwise_df.R
llrs-roche: Yeah, I see the SE is big and so the CI is also enormous but what's is also surprising is that is the only variable it happens, and coincidence or not is the last variable. Sorry, I should have pointed that out also on the initial comment.

R/get_cox_pairwise_df.R
llrs-roche: Ah, I see, so the warning has moved up. Thanks for clarifying.

R/get_cox_pairwise_df.R
llrs-roche: If you have a `@keywords internal` help page it won't show up on the manual, but users can read it if they explore the help pages for example using `help(package = "crane")`.

CRAN lately has been checking that all available help pages have Examples and Values section with relevant content. I think it might be hard to come up with an example that is useful to users that's why I said to use `@noRd`. Maybe this manual review of help pages is not done as {crane} won't be a new submission but I hope that with this we avoid a delay on the acceptance of the package.

R/get_cox_pairwise_df.R
llrs-roche: It is kind of explained when the output is assigned to wald_ci_log but indeed, a comment won't hurt: I wasn't sure if Wald's CI was calculated by the confint.glm method or the default.

R/get_cox_pairwise_df.R
jszczypinski: The code will now work for formulas with covariates and for formulast with `strata(covariate)`

tests/testthat/test-get_cox_pairwise_df.R
jszczypinski: Done :)

R/get_cox_pairwise_df.R
jszczypinski: I added info to docs that Wald CIs are calculated and also another comment in the main function.

R/get_cox_pairwise_df.R
Melkiades: I agree with you. It has created many false positive warnings, often for dev only function descriptions. We should probably harmonize this, by using `@noRd` where examples are not or would not be meaningful

Melkiades: But for another time

Melkiades: maybe open an issue to track it ;)

jszczypinski: Tracked #257

R/get_cox_pairwise_df.R
jszczypinski: No longer relevant. We will discard this part of the code

jszczypinski:
I am closing this PR so that the naming and the actual changes matters. Changes to formulas and new feature to include `strata()` in the formulas are tracked here #256

PR #258, opened 2026-05-27
256 get cox pairwise df strata in formulas

### LINKED ISSUE #256: Make `get_cox_pairwise_df` accept formulas with `strata()` and remove namespace from formulas
Currently `get_cox_pairwise_df` do not accept formulas with `strata()` term, which is quite common.
Also formulas in R should not use namespace 
https://stat.ethz.ch/pipermail/r-devel/2025-April/083989.html

jszczypinski:
Currently get_cox_pairwise_df do not accept formulas with strata() term, which is quite common.
Also formulas in R should not use namespace
https://stat.ethz.ch/pipermail/r-devel/2025-April/083989.html

Closes #256 

- [ ] Request a reviewer

- [ ] Approve Pull Request


R/get_cox_pairwise_df.R
```diff
@@ -44,28 +45,29 @@
 #'   data to the current arm and the reference arm, and then:
 #'   \itemize{
 #'     \item Fits a Cox model using `survival::coxph()`.
-#'     \item Computes a p-value via [.estimate_p_value()], which dispatches
+#'     \item Computes a p-value via, which dispatches
```
jszczypinski:
```suggestion
#'     \item Computes a p-value, which dispatches
```

R/get_cox_pairwise_df.R
```diff
@@ -44,28 +45,29 @@
...
 #'       to `coin::logrank_test()` for weighted log-rank variants or to
 #'       `survival::survreg()` for the likelihood-ratio test.
-#'   }
+#' }
 #'
 #' @seealso `annotate_gg_km()`, `gg_km()`, `survival::coxph()`,
 #'   `coin::logrank_test()`.
 #'
-#' @examples
+#' @examplesIf requireNamespace("survival", quietly = TRUE)
```
jszczypinski:
```suggestion
#' @examplesIf requireNamespace("survival", quietly = TRUE) && requireNamespace("coin", quietly = TRUE)
```

tests/testthat/test-get_cox_pairwise_df.R
```diff
@@ -1,30 +1,28 @@
 skip_if_pkg_not_installed(c("survival", "dplyr", "coin"))
 
-# Setup shared test data
+# Setup shared test data using completely different data and formulas
```
jszczypinski:
```suggestion
# Setup shared test data 
```

R/get_cox_pairwise_df.R
```diff
@@ -217,43 +232,168 @@ get_cox_pairwise_df <- function(
...
+    reduced_formula <- stats::update(formula, paste(". ~ . -", arm))
+
+    # Fit the reduced model explicitly to avoid drop1 environment scope errors
+    fit_reduced <- survival::coxph(reduced_formula, data = data, ties = ties)
+
+    # Execute the nested Likelihood-Ratio Test
+    anova_res <- stats::anova(fit_reduced, fit_full, test = "Chisq")
+
+    # Extract the p-value for the second row (the full model comparison)
+    p_value <- anova_res[["Pr(>|Chi|)"]][[2]]
```
jszczypinski: I have re-written this part to simplify it by using `anova()` and so that `strata()` in formula can be evaluated natively

R/annotate_gg_km.R
```diff
@@ -46,15 +46,16 @@
 #' @seealso [gg_km()], [process_survfit()], and [get_cox_pairwise_df()] for
 #'   related functionalities.
 #'
-#' @examples
+#' @examplesIf requireNamespace("survival", quietly = TRUE) && requireNamespace("coin", quietly = TRUE)
```
Melkiades: as Lluis said, if we have it in import it is not really relevant to have it (coin is in suggests right?)

R/get_cox_pairwise_df.R
```diff
@@ -35,7 +36,7 @@
 #' The columns are:
 #' \itemize{
 #'   \item `HR`: The Hazard Ratio formatted to two decimal places.
-#'   \item `95% CI`: The 95\% confidence interval as `"(lower, upper)"`.
+#'   \item `95% CI`: The 95% Wald confidence interval as `"(lower, upper)"`.
```
Melkiades: can you change the method if you want? If not we need to be upfront about it

R/get_cox_pairwise_df.R
```diff
@@ -116,13 +118,23 @@ get_cox_pairwise_df <- function(
...
   }
+
+  formula_str <- paste(deparse(model_formula), collapse = " ")
+  if (grepl("\\b[a-zA-Z][a-zA-Z0-9.]*::", formula_str)) {
+    cli::cli_abort(
+      "{.arg model_formula} must be specified without namespace.",
+      call = get_cli_abort_call()
+    )
+  }
+
```
Melkiades: I got the same from AI in another PR (avot01). I wonder if we should make an utility for it

R/get_cox_pairwise_df.R
jszczypinski: By default, `coxph()` calculates the variance using the standard Fisher information matrix, and the resulting CIs extracted via `summary()` or `confint()` are Wald confidence intervals.

If you want to change the underlying variance method to calculate Robust (Sandwich) Confidence Intervals, you need to use the `robust = TRUE` argument inside your model call.

So we can add an argument `robust` and additionally specify the confidence limit in another argument?

However these will be two additional arguments in the function call. What is your take on this?

R/get_cox_pairwise_df.R
jszczypinski: I will put it `import-standalone-checks.R`

R/annotate_gg_km.R
jszczypinski: I will change this. coin is in suggests and I thought survival is also in suggests

R/get_cox_pairwise_df.R
llrs-roche: Could we use `...` to pass these additional arguments consistently to each method?

R/get_cox_pairwise_df.R
jszczypinski: It's called `check_formula_for_namespace()`

Melkiades: That is not the place for this. see header:
```r
# Standalone file: do not edit by hand
# Source: https://github.com/insightsengineering/standalone/blob/HEAD/R/standalone-checks.R
# Generated by: usethis::use_standalone("insightsengineering/standalone", "checks")
# ----------------------------------------------------------------------
```

R/import-standalone-checks.R
```diff
@@ -604,5 +604,18 @@ check_numeric <- function(x,
   invisible(x)
 }
 
+#' Check if formula contains namespace
+#' @keywords internal
+#' @noRd
+check_formula_for_namespace <- function(formula) {
```
Melkiades: I think we need a test for this directly (to add into utils.R)

tests/testthat/test-get_cox_pairwise_df.R
```diff
@@ -40,18 +38,20 @@ test_that(
     expect_no_error(
       suppressWarnings(
```
Melkiades: I would aim at removing all suppressWarnings (I may also have added some tbh)

R/get_cox_pairwise_df.R
jszczypinski: Good call. I have put it in `utils.R`

R/get_cox_pairwise_df.R
jszczypinski: I have added the code to do that.

R/get_cox_pairwise_df.R
```diff
@@ -187,73 +243,143 @@ get_cox_pairwise_df <- function(
...
+  } else if (test == "likelihood-ratio") {
+    # --- 2. Likelihood-Ratio Test via survival ---
+    # Fit the full Cox model
+    fit_full <- survival::coxph(formula, data = data, ties = ties)
+
+    # Safely create the reduced formula by dropping the arm variable
+    reduced_formula <- stats::update(formula, paste(". ~ . -", arm))
+
+    # Fit the reduced model explicitly to avoid drop1 environment scope errors
+    fit_reduced <- survival::coxph(reduced_formula, data = data, ties = ties)
```
jszczypinski: I have reverted to survival for computing log-rank p value, since coin is using permutation test for this. This resulted in discrepancy between crane and tern results.

R/import-standalone-checks.R
jszczypinski: Added a test for it

tests/testthat/test-get_cox_pairwise_df.R
jszczypinski: I removed all from `get_cox_pairwise_df()`

R/get_cox_pairwise_df.R
Melkiades: We need to talk about the LRT engine change (`survreg` → `coxph`)

Just flagging this because I think it deserves a note in the PR description / NEWS.

The old LRT used `survreg(dist = "exponential")` to match SAS's `PROC LIFETEST` LR test. The new code uses nested `coxph()` anova (Cox partial-likelihood LRT, equivalent to SAS `PROC PHREG`). The switch was necessary because `survreg()` errors on `strata()` terms.

However, we could preserve both: use the original `survreg` exponential LRT when the formula has no `strata()` terms (keeping backward compatibility and SAS `PROC LIFETEST` alignment), and fall back to the Cox LRT only when `strata()` is present. This way existing non-stratified results stay unchanged, and stratified formulas get a valid test. Both approaches have SAS equivalents — they just come from different procedures.

Either way, worth a note in NEWS since `test = "likelihood-ratio"` behavior is changing for at least the stratified case.

jszczypinski: That's a good point. Lets do as you said. Use survreg if there is no strata and coxph if there is strata in formula. I will also add a warning about it so the users get notified

Melkiades (commented):
Nice work on the dual LRT engine, the `...` passthrough, and switching log-rank to `survdiff`. A few remaining items:

### 1. Broken examples in `annotate_gg_km` and `gg_km` 

The `@examplesIf` on the `@name` block still closes before the `@describeIn` `@examples` that depend on objects created inside it. In the generated `.Rd`:

\dontshow{if (...) withAutoprint({ # examplesIf}

... creates plt_kmg01, fit_kmg01 ...
\dontshow{}) # examplesIf}

Runs unconditionally — plt_kmg01 doesn't exist here:
annotate_riskdf(plt_kmg01, fit_kmg01)


Same issue in `gg_km.Rd`. Either wrap all examples in a single `@examplesIf`, or add the guard to each `@describeIn` block.

### 2. Coverage at 69% on `get_cox_pairwise_df.R` (−24% vs main)

The CI bot reports 50 uncovered lines. The main gaps are:

- **Lines 140–147**: invalid `...` args error — needs a test like `expect_error(get_cox_pairwise_df(..., bad_arg = 1), "Unrecognized")`
- **Lines 172–178**: `robust = TRUE` + `ties = "exact"` error — needs a test
- **Lines 317–322**: `coin` not installed abort — can skip this one (hard to test without mocking)
- **Lines 365–409**: `.rewrite_formula()` guardrail for covariates in coin tests — the existing strata tests don't hit the `invalid_terms` error path. Add a test with e.g. `Surv(time, event) ~ treatment + karno` and `test = "peto"` to cover it.



### 3. `@param ...` doc says "two arguments are supported" but doesn't mention `conf.int` changes the column name

The CI column name is now dynamic (`paste0(conf_int * 100, "% CI")`), which is a nice touch. But the `@return` doc still says `"95% CI"` — it should mention the column name reflects the chosen level. Also, downstream code that does `result[["95% CI"]]` will break if someone passes `conf.int = 0.99`. Worth a note.



### 4. `robust` is not passed to `.estimate_p_value()` for the LRT path

In `get_cox_pairwise_df()`, `robust` is injected into `coxph_dots` and passed to the main `coxph` call via `do.call`. But `.estimate_p_value()` fits its own `coxph` models for the LRT and doesn't receive `robust`. So the HR/CI use robust SEs but the LRT p-value doesn't — this is inconsistent. Either pass `robust` (and `coxph_dots`) through to `.estimate_p_value()`, or document that the LRT always uses model-based SEs.



### 5. `.check_formula_for_namespace()` error message says `{.arg formula}` but the user-facing param is `model_formula`

In `utils.R`:
```r
cli::cli_abort("{.arg formula} must be specified without namespace.")

File: R/get_cox_pairwise_df.R [hunk 11]
```diff
@@ -187,73 +243,168 @@ get_cox_pairwise_df <- function(
...
+    if (!requireNamespace("coin", quietly = TRUE)) {
+      cli::cli_abort(
         paste(
-          "{.arg arm} does not contain exactly 2 levels!",
-          "This will result in unexpected behavior of pairwise test."
+          "The {.pkg coin} package is required to run",
+          "log-rank tests using the {test_type} method."
         )
       )
     }
```
llrs-roche: Maybe we could reuse `rlang::check_installed("coin")`

llrs-roche (approved):
Beyond @Melkiades comments,  I'll add one very minor comment, and one bigger issue:

```
E  checking examples ... [77s] ERROR (1m 18.9s)
   Running examples in 'crane-Ex.R' failed
   The error most likely occurred in:
   
   > base::assign(".ptime", proc.time(), pos = "CheckExEnv")
   > ### Name: annotate_gg_km
   > ### Title: Annotate Kaplan-Meier Plot
   > ### Aliases: annotate_gg_km annotate_riskdf annotate_surv_med
   > ###   annotate_coxph
   > 
   > ### ** Examples
   > 
   > ## Don't show: 
   > if (requireNamespace("coin", quietly = TRUE)) withAutoprint({ # examplesIf
   + ## End(Don't show)
   + # Preparing the Kaplan-Meier Plot
   + library(survival)
   + use_lung <- lung
   + use_lung$arm <- factor(sample(c("A", "B", "C"), nrow(use_lung), replace = TRUE))
   + use_lung$status <- use_lung$status - 1 # Convert status to 0/1
   + use_lung <- na.omit(use_lung)
   + 
   + formula <- Surv(time, status) ~ arm
   + fit_kmg01 <- survfit(formula, use_lung)
   + surv_plot_data <- process_survfit(fit_kmg01)
   + 
   + plt_kmg01 <- gg_km(surv_plot_data)
   + ## Don't show: 
   + }) # examplesIf
   > ## End(Don't show)
   > # Annotate Plot with Numbers at Risk Table
   > annotate_riskdf(plt_kmg01, fit_kmg01)
   Error: object 'plt_kmg01' not found
   Execution halted
```

Test look great! Thanks for providing `skip_if_not_installed` on some files. 

(I saw 31 test fail, but I believe this is because I have some development branch installed)

R/get_cox_pairwise_df.R
jszczypinski: I wanted to be consistent and always use cli package for warnings/errors in new code.

R/get_cox_pairwise_df.R
```diff
@@ -313,14 +321,10 @@ get_cox_pairwise_df <- function(
...
-        paste(
-          "The {.pkg coin} package is required to run",
-          "log-rank tests using the {test_type} method."
-        )
-      )
-    }
+    rlang::check_installed(
+      "coin",
+      reason = paste("to run log-rank tests using the", test_type, "method.")
+    )
```
Melkiades, line 329: very nice!

R/get_cox_pairwise_df.R
jszczypinski: However as you see in comment below I have returned to `rlang::check_installed("coin")` :)
It was much easier to test it

R/get_cox_pairwise_df.R
jszczypinski: Based on @llrs-roche comment

R/get_cox_pairwise_df.R
```diff
@@ -187,48 +251,83 @@ get_cox_pairwise_df <- function(
...
+    sdiff <- survival::survdiff(formula, data = data)
+
+    # Calculate the asymptotic p-value using the chi-square statistic
+    # Degrees of freedom is (number of groups - 1)
+    p_value <- stats::pchisq(sdiff$chisq, df = length(sdiff$n) - 1, lower.tail = FALSE)
+  } else if (test == "likelihood-ratio") {
+    # --- 2. Likelihood-Ratio Test via survival ---
+    if (.check_has_strata(formula)) {
+      # Inform user of the engine swap for stratified LRTs
+      cli::cli_warn(
```
Melkiades, line 286: Missing `call =` in internal helpers' `cli_abort`/`cli_warn`

The exported `get_cox_pairwise_df()` correctly uses `set_cli_abort_call()` + `call = get_cli_abort_call()` on all its `cli_abort` calls. But the internal helpers don't propagate this:

- `.estimate_p_value()` line ~286: `cli::cli_warn(...)` — no `call =`
- `.rewrite_formula()` line ~381: `cli::cli_abort(...)` — no `call =`

Without `call =`, the error/warning backtrace will point to the internal helper instead of the user-facing `tbl_ancova()` or `get_cox_pairwise_df()` call. Fix: pass `call = get_cli_abort_call()` in both places, or accept `call` as a parameter from the parent.

R/annotate_gg_km.R
```diff
@@ -48,13 +48,14 @@
 #'
 #' @examples
```
Melkiades: We need to open an issue here as if no survival is installed these break (as exampleif should be in all the page)

R/get_cox_pairwise_df.R
```diff
@@ -26,16 +26,30 @@
...
+#' computed using robust sandwich standard errors. However, the p-values across all
+#' tests (including the likelihood-ratio test) are calculated using standard,
+#' non-robust model variances.
 #'
 #' @return A `data.frame` with one row per comparison arm (stored as rownames).
 #' The columns are:
 #' \itemize{
 #'   \item `HR`: The Hazard Ratio formatted to two decimal places.
-#'   \item `95% CI`: The 95\% confidence interval as `"(lower, upper)"`.
+#'    \item `conf.int` (default `0.95`): Adjusts the confidence interval level.
```
Melkiades:
```suggestion
#'   \item `conf.int` (default `0.95`): Adjusts the confidence interval level.
```

Melkiades: I know I asked for this but it is still a bit confusing to have that as column name... Maybe there is a better way to show it correctly with a placeholder? or just highlighting the note afterwards

Melkiades (commented):
Almost good to go!! Sorry Jan for so many comments. remember the news: e.g.

- Strata support in `get_cox_pairwise_df()`
- `...` passthrough (`conf.int`, `robust`)
- Namespace rejection
- Log-rank engine change from `coin` to `survdiff` for standard log-rank
- Dual LRT engine (survreg for non-stratified, coxph for stratified)

These are user-facing behavioral changes and need entries.

tests/testthat/test-get_cox_pairwise_df.R
```diff
@@ -1,30 +1,30 @@
 skip_if_pkg_not_installed(c("survival", "dplyr", "coin"))
+suppressPackageStartupMessages(library(survival))
```
Melkiades: `withr::local_package("survival")` is better! library calls risk to load it for the whole testing session slowing it down

R/annotate_gg_km.R
jszczypinski: I thought that if survival is in Imports then it should be good without @ExamplesIf. I will add it so everything will work on CRAN

Melkiades: yep that would work

R/get_cox_pairwise_df.R
```diff
@@ -66,7 +66,7 @@
 #' @seealso `annotate_gg_km()`, `gg_km()`, `survival::coxph()`,
 #'   `coin::logrank_test()`.
 #'
-#' @examplesIf requireNamespace("coin", quietly = TRUE)
+#' @examplesIf requireNamespace("coin", quietly = TRUE) && requireNamespace("survival", quietly = TRUE) && requireNamespace("dplyr", quietly = TRUE)
```
Melkiades, line 69: if they are in import we do not need it imo

R/get_cox_pairwise_df.R
```diff
@@ -271,7 +271,7 @@ get_cox_pairwise_df <- function(
 #'
 #' @returns A single numeric p-value.
 #' @noRd
-.estimate_p_value <- function(formula, data, test, arm, ties) {
+.estimate_p_value <- function(formula, data, test, arm, ties, call = get_cli_abort_call()) {
```
Melkiades: I think you just need to add get_cli_abort_call() to the call = . It should use env to find it or something like this. No need of an extra param

R/get_cox_pairwise_df.R
jszczypinski: They are in import, but I have put them there anyway to be safe. Lets keep it for now and change it in future development if needed.

PR #265, opened 2026-06-11
feat: expose `overall_row`, `add_overall()`, and spanning headers for `tbl_hierarchical_incidence_rate()`

### LINKED ISSUE #264: Expose `overall_row` and `add_overall()` for `tbl_hierarchical_incidence_rate()`
`tbl_hierarchical_incidence_rate()` has three gaps:

1. `overall_row` is hardcoded to `TRUE` -- users cannot opt out of the overall summary row.
2. No `add_overall()` S3 method -- users cannot append an unstratified "Overall" column pooling all treatment arms.
3. No spanning headers -- the treatment arm labels are missing above the column groups after `tbl_merge()`.

This PR:
- Exposes `overall_row` as an explicit argument (default `TRUE`).
- Adds `add_overall.tbl_hierarchical_incidence_rate()` that rebuilds the table with `by = NULL` and merges the overall columns.
- Adds spanning headers for treatment arm column groups.

Melkiades:
* `tbl_hierarchical_incidence_rate()` gains an `overall_row` argument to control whether the overall summary row is included. (#264, @Melkiades)
* `tbl_hierarchical_incidence_rate()` now supports `add_overall()` to append an unstratified "Overall" column pooling all treatment arms. (#264, @Melkiades)
* `tbl_hierarchical_incidence_rate()` now renders spanning headers for treatment arm columns. (#264, @Melkiades)

The `add_overall()` method rebuilds the entire table with `by = NULL` and merges the resulting 5 panels (count, n_events, PY, rate, CI) as `stat_0_*` columns into the existing table. Spanning headers use `tidyselect::starts_with()` formulas to group columns by arm.

**Reference GitHub issue associated with pull request.** closes #264

- [ ] Request a reviewer

### DIFF
--- a/R/tbl_hierarchical_incidence_rate.R
+++ b/R/tbl_hierarchical_incidence_rate.R
@@ -397,10 +397,11 @@ tbl_hierarchical_incidence_rate <- function(data,
 #' @rdname tbl_hierarchical_incidence_rate
 #' @export
 add_overall.tbl_hierarchical_incidence_rate <- function(
-    x,
-    last = FALSE,
-    col_label = "All Participants  \n(N = {style_roche_number(N)})",
-    ...) {
+  x,
+  last = FALSE,
+  col_label = "All Participants  \n(N = {style_roche_number(N)})",
+  ...
+) {
   # check inputs ---------------------------------------------------------------
   set_cli_abort_call()
   check_dots_empty(call = get_cli_abort_call())
--- a/R/tbl_hierarchical_incidence_rate.R
+++ b/R/tbl_hierarchical_incidence_rate.R
@@ -48,6 +48,21 @@
 #'   Whether to include an overall summary row aggregating across all
 #'   hierarchical levels. Default is `TRUE`. The overall row label can be
 #'   customized via `label = list("..ard_hierarchical_overall.." = "Custom Label")`.
+#' @param spanning_label (`string`)\cr
+#'   A [glue][glue::glue]-style template for the per-arm spanning headers when
+#'   `by` is supplied. Available substitutions are `{level}` (the arm level) and
+#'   `{n}` (the arm participant count). Default is `"{level}  \\n(N = {n})"`,
+#'   matching the column-header convention used elsewhere in the package.
+#' @param x (`tbl_hierarchical_incidence_rate`)\cr
+#'   A table created with [tbl_hierarchical_incidence_rate()].
+#' @param last (`logical(1)`)\cr
+#'   When `add_overall()` is used, whether to place the overall columns after
+#'   (`TRUE`) or before (`FALSE`) the stratified columns. Default is `FALSE`.
+#' @param col_label (`string`)\cr
+#'   A [glue][glue::glue]-style template for the overall spanning header added by
+#'   `add_overall()`. Resolved against the column header context (e.g. `{N}`,
+#'   `{n}`). Default is `"All Participants  \\n(N = {style_roche_number(N)})"`.
+#' @param ... These dots are for future extensions and must be empty.
 #'
 #' @returns a gtsummary table of class `"tbl_hierarchical_incidence_rate"`.
 #' @name tbl_hierarchical_incidence_rate
@@ -126,7 +141,7 @@
 #' ) |>
 #'   add_overall()
 #'
-#' # Customize spanning headers after table creation
+#' # Customize the per-arm spanning headers with a glue template
 #' tbl_hierarchical_incidence_rate(
 #'   data = adae,
 #'   denominator = adsl,
@@ -134,12 +149,9 @@
 #'   by = ARM,
 #'   start_date = TRTSDT,
 #'   end_date = TRTEDT,
-#'   event_date = AESTDTC
-#' ) |>
-#'   gtsummary::modify_spanning_header(
-#'     tidyselect::starts_with("stat_1") ~ "Placebo (N = 3)",
-#'     tidyselect::starts_with("stat_2") ~ "Treatment (N = 2)"
-#'   )
+#'   event_date = AESTDTC,
+#'   spanning_label = "{level} (N = {n})"
+#' )
 #'
 #' @export
 tbl_hierarchical_incidence_rate <- function(data,
@@ -157,7 +169,9 @@ tbl_hierarchical_incidence_rate <- function(data,
                                             conf.type = "normal",
                                             digits = 2,
                                             overall_row = TRUE,
+                                            spanning_label = "{level}  \n(N = {n})",
                                             label = NULL) {
+  set_cli_abort_call()
   # 1. Base Input Validation (Missingness and standard classes)
   event_type <- rlang::arg_match(event_type)
 
@@ -177,8 +191,9 @@ tbl_hierarchical_incidence_rate <- function(data,
   check_numeric(digits)
   check_scalar(digits)
   check_scalar_logical(overall_row)
+  check_string(spanning_label)
   if (!is.null(label) && !is.list(label)) {
-    cli::cli_abort("{.arg label} must be a list or NULL.")
+    cli::cli_abort("{.arg label} must be a list or NULL.", call = get_cli_abort_call())
   }
 
   # 2. Process Tidy-Select Arguments (Evaluates unquoted inputs to strings)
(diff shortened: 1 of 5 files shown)


tests/testthat/test-tbl_hierarchical_incidence_rate.R
```diff
@@ -133,6 +133,7 @@ test_that("tbl_hierarchical_incidence_rate fails safely on bad inputs", {
 
   # 2. Invalid Dataframes
   expect_error(
+
```
Melkiades:
```suggestion
```

R/tbl_hierarchical_incidence_rate.R
```diff
@@ -171,28 +214,34 @@ tbl_hierarchical_incidence_rate <- function(data,
...
       gtsummary::all_stat_cols() ~ "No. of\nParticipants\nwith AE (%)"
+    )
+
+  if (overall_row) {
+    tbl_base <- tbl_base |>
+      gtsummary::modify_table_body(~ .x |> dplyr::mutate(
+        label = dplyr::if_else(
+          dplyr::row_number() == 1, .env$overall_label, .data$label
+        )
+      ))
```
Melkiades, line 259: why this?

R/tbl_hierarchical_incidence_rate.R
```diff
@@ -277,11 +334,123 @@ tbl_hierarchical_incidence_rate <- function(data,
...
+        lhs = rlang::expr(tidyselect::starts_with(!!prefix)),
+        rhs = rlang::expr(!!spanner_label)
+      )
+    })
+
+    tbl_final <- rlang::inject(
+      gtsummary::modify_spanning_header(tbl_final, !!!spanning_formulas)
+    )
+  }
+
```
Melkiades: How easy it would be to change this lets say to Placebo (N =3)

R/tbl_hierarchical_incidence_rate.R
```diff
@@ -277,11 +334,123 @@ tbl_hierarchical_incidence_rate <- function(data,
...
+  tbl_final$call_list <- list(tbl_hierarchical_incidence_rate = match.call())
+
   class(tbl_final) <- c("tbl_hierarchical_incidence_rate", class(tbl_final))
 
   tbl_final
 }
 
+#' @rdname tbl_hierarchical_incidence_rate
+#' @export
+add_overall.tbl_hierarchical_incidence_rate <- function(
```
Melkiades, line 399: is this aligned with classical add_overall() functions from gtsummary? is the ard similarly formatted?

tests/testthat/test-tbl_hierarchical_incidence_rate.R
Melkiades: Fixed.

R/tbl_hierarchical_incidence_rate.R
Melkiades: `tbl_hierarchical(overall_row = TRUE)` generates a default label for the overall row. This overrides it with the user-specified label from `label = list("..ard_hierarchical_overall.." = "...")`. Without this, the custom label would be ignored.

R/tbl_hierarchical_incidence_rate.R
Melkiades: Trivial -- it is just the `paste0` on this line. We could also expose it as a `col_label` style argument with glue syntax (e.g. `spanning_label = "{level} (N = {n})"`) if we want it user-configurable. For now I kept it simple. Let me know if you want a different default or a parameter.

R/tbl_hierarchical_incidence_rate.R
Melkiades: Not directly -- the other crane `add_overall` methods (e.g. `tbl_hierarchical_rate_and_count`) delegate to `gtsummary:::add_overall.tbl_hierarchical` via `do.call`. That works because those are single `tbl_hierarchical` objects. Here the table is a `tbl_merge` of 5 panels, so gtsummary's `add_overall_merge()` breaks (it expects a single `stat_0` column). Instead we rebuild the full table with `by = NULL` and bind the 5 resulting columns manually. The ARD is not stored in the same way -- we skip `x$cards[["add_overall"]]` since the merged structure does not have a single cards slot. If we need ARD extraction from the overall columns later we can add it.

R/tbl_hierarchical_incidence_rate.R
Melkiades: I mean from an user perspective in post-processing?

R/tbl_hierarchical_incidence_rate.R
Melkiades: fine for the custom add_overall method but shouldnt we respect the usual-defined ard structure?

R/tbl_hierarchical_incidence_rate.R
Melkiades: Good point. Users can already call `modify_spanning_header()` after the fact to override, but having to know the column naming convention (`stat_1_*`, `stat_2_*`) is not great. I can add a `spanning_label` argument with glue support -- something like `spanning_label = "{level}\nN = {n}"` so users can do `spanning_label = "{level} (N = {n})"` without touching internals. Want me to add that?

R/tbl_hierarchical_incidence_rate.R
Melkiades: Yes, we should. Right now the overall ARD lives inside the rebuilt table but is not propagated to `x$cards`. I can store it as `x$cards[["add_overall"]] <- tbl_overall$cards[[1]]` to keep it consistent with how gtsummary does it. The tricky part is that we have multiple cards slots from the merge -- I will map the overall cards into the right slot so `gather_ard()` picks them up correctly.

Melkiades: Done. The overall table's inner `tbls` are now appended to `x$tbls` so `gather_ard()` recurses into them. Added a test confirming `gather_ard()` returns more elements after `add_overall()`.

R/tbl_hierarchical_incidence_rate.R
```diff
@@ -143,15 +200,16 @@ tbl_hierarchical_incidence_rate <- function(data,
...
-  tbl_final$call_list <- list(tbl_hierarchical_incidence_rate = match.call())
-  tbl_final$inputs <- as.list(environment())
 
-  # 4. Extract overall label BEFORE `cards` processes the dataset
-  overall_label <- "All Adverse Events"
-  if (is.list(label) && "..ard_hierarchical_overall.." %in% names(label)) {
-    overall_label <- label[["..ard_hierarchical_overall.."]]
+  # Extract overall label BEFORE `cards` processes the dataset
+
+  overall_label <- NULL
```
llrs-roche: Do we need to define it as NULL? if it is required by default it will become "All Adverse Events" as defined two lines below

R/tbl_hierarchical_incidence_rate.R
```diff
@@ -171,28 +229,34 @@ tbl_hierarchical_incidence_rate <- function(data,
...
+    tbl_base <- tbl_base |>
+      gtsummary::modify_table_body(~ .x |> dplyr::mutate(
+        label = dplyr::if_else(
+          dplyr::row_number() == 1, .env$overall_label, .data$label
+        )
+      ))
+  }
+
+  # Generate ARDs cleanly using explicit piping
+  if (overall_row) {
```
llrs-roche: This code block starts after the previous code block also starts with `if (overall_row) {`, they can be joined together.

llrs-roche (commented):
Many times there are calls for `grepl("^stat_", .data$column)` or `grep("^stat_", names(..), value = TRUE). Maybe a helper is needed? They are mostly used to select and order columns.

Related the `add_overall()` method seems very different for the different classes...

Looks like a fine addition, but I am a bit confused by the choice of the approach, why having a new argument on the middle instead of a new function or the argument on the end. This last approach would be compatible with positional based arguments and might break less user's code. 

I might have more feedback when I'm more fresh

llrs-roche (approved):
I don't see anything bad. But there are two comments unanswered: one about why the new arguments are on that position and the other one about having helpers to operate with stat_ variables.

Melkiades:
> I don't see anything bad. But there are two comments unanswered: one about why the new arguments are on that position and the other one about having helpers to operate with stat_ variables.

Thank you for the review Lluis. Apologies to not have responded you on the two points above. Took a moment to consider them and


For the first point I followed gtsummary default orders as it is the point of reference (tbl_hierarchical() and add_overall() there). It is an exception spanning_label that works well with glue like normal label imo.

For the stat_, I think it is quite readable and a wrapper would only consume more lines of code for the function itself

PR #274, opened 2026-06-21
Frame flextable header column labels with an outer border only

### LINKED ISSUE #272: theme_gtsummary_roche() draws a vertical frame around the header
## Problem

`theme_gtsummary_roche()` applies `flextable::border_outer(part = "header")`, which draws all four borders around the header part. In rendered `.docx` tables this produces a box around the title/study/parameter header block, and because of how `border_outer` interacts with the column-label row it leaves an inconsistent missing right border on the last header row.

TLG outputs use horizontal rules only; there should be no vertical frame around the header.

## Expected

Header keeps a top rule, a bottom rule and inner horizontal rules between header rows, with no left/right (vertical) borders.

## Fix

Replace `border_outer(part = "header")` with `hline_top()` + `hline_bottom()` on the header part, keeping the existing `border_inner_h(part = "header")`. Cosmetic, no change in principle; the only test impact is regenerating the theme command snapshot.

Melkiades:
* `theme_gtsummary_roche()` now frames the flextable column labels with an outer border only, removing the internal borders between header rows and the inconsistent missing right border. (#272)

`border_inner_h(part = "header")` clears the rules between header rows, then `border_outer(part = "header")` draws a single box around the column-label block. The inner border uses `officer::fp_border(width = 0, style = "none")` so docx does not render a line between a spanning header and the column labels (a width-0 *solid* border is still written as a visible line). The theme command snapshot is regenerated accordingly.

**Reference GitHub issue associated with pull request.**
closes #272

### DIFF
--- a/R/theme_gtsummary_roche.R
+++ b/R/theme_gtsummary_roche.R
@@ -96,7 +96,12 @@ theme_gtsummary_roche <- function(font_size = NULL,
         ),
         valign = list( # valign only because it will append to to last commands
           rlang::expr(flextable::fontsize(size = !!((font_size %||% 8) - 1), part = "footer")), # second fontsize spectbl
-          rlang::expr(flextable::border_outer(part = "header", border = !!border)), # second command from border
+          # Horizontal rules only on the header: a top and bottom rule plus the
+          # inner horizontal rules between header rows. Avoid border_outer() here
+          # so no vertical frame is drawn around the title/study/parameter block
+          # (which also left an inconsistent missing right border).
+          rlang::expr(flextable::hline_top(part = "header", border = !!border)),
+          rlang::expr(flextable::hline_bottom(part = "header", border = !!border)),
           rlang::expr(flextable::border_inner_h(part = "header", border = !!border)), # fix different line sizes
           rlang::expr(flextable::valign(valign = "top", part = "all")),
           rlang::expr(flextable::font(fontname = "Arial", part = "all")),
--- a/R/theme_gtsummary_roche.R
+++ b/R/theme_gtsummary_roche.R
@@ -96,11 +96,11 @@ theme_gtsummary_roche <- function(font_size = NULL,
         ),
         valign = list( # valign only because it will append to to last commands
           rlang::expr(flextable::fontsize(size = !!((font_size %||% 8) - 1), part = "footer")), # second fontsize spectbl
-          # Column labels get an outer border only: clear any internal borders
-          # between header rows, then frame the block with a single outer box.
+          # outer box only. style = "none" (not width = 0) so docx doesn't draw
+          # a line between a spanner and the column labels.
           rlang::expr(flextable::border_inner_h(
             part = "header",
-            border = flextable::fp_border_default(width = 0)
+            border = officer::fp_border(width = 0, style = "none")
           )),
           rlang::expr(flextable::border_outer(part = "header", border = !!border)),
           rlang::expr(flextable::valign(valign = "top", part = "all")),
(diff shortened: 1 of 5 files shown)


R/theme_gtsummary_roche.R
```diff
@@ -96,7 +96,12 @@ theme_gtsummary_roche <- function(font_size = NULL,
         ),
         valign = list( # valign only because it will append to to last commands
           rlang::expr(flextable::fontsize(size = !!((font_size %||% 8) - 1), part = "footer")), # second fontsize spectbl
-          rlang::expr(flextable::border_outer(part = "header", border = !!border)), # second command from border
+          # Horizontal rules only on the header: a top and bottom rule plus the
+          # inner horizontal rules between header rows. Avoid border_outer() here
+          # so no vertical frame is drawn around the title/study/parameter block
+          # (which also left an inconsistent missing right border).
+          rlang::expr(flextable::hline_top(part = "header", border = !!border)),
+          rlang::expr(flextable::hline_bottom(part = "header", border = !!border)),
```
Melkiades: is this keeping header column labels borders? no internal borders only outer only for column labels.

shajoezhu:
let's wait this a little bit for feature confirmation.

R/theme_gtsummary_roche.R
Melkiades: Yes. The header (column-label block) now gets an outer border only: I clear the inner horizontal borders with `border_inner_h(part = "header", border = fp_border_default(width = 0))` and then apply `border_outer(part = "header")`. So a single box frames the column labels with no internal rules between header rows.

Verified in a rendered docx with a spanning header (two header rows): top rule on the first row, bottom rule on the last, left/right on both, and no border between them.

Note: title/subtitle rows are added downstream by citril, where their vertical frame is stripped — so the box stays scoped to the column labels.

tests/testthat/test-theme_gtsummary_roche.R
```diff
@@ -119,8 +119,60 @@ test_that("theme pre-conversion modifies header not to be bold and border only 0
...
+        gtsummary::as_flex_table()
+    }
+  )
+
+  f <- withr::local_tempfile(fileext = ".docx")
+  flextable::save_as_docx(tbl, path = f)
+  d <- withr::local_tempdir()
+  utils::unzip(f, exdir = d)
+  xml <- paste(readLines(file.path(d, "word", "document.xml"), warn = FALSE), collapse = "")
+  rows <- regmatches(xml, gregexpr("<w:tr\\b.*?</w:tr>", xml))[[1]]
```
llrs-roche: I find it odd that the package is using regular expressions for xml checks. It will be much easier with xml2 package and using Xpath to find the right elements and attributes.

Melkiades: I think here we do not have xml2 as dep

Melkiades: Switched to xml2/XPath it's an indirect dep of officer/flextable which this test already requires, so no new dependency

tests/testthat/test-theme_gtsummary_roche.R
```diff
@@ -119,8 +119,73 @@ test_that("theme pre-conversion modifies header not to be bold and border only 0
...
+  flextable::save_as_docx(tbl, path = f)
+  d <- withr::local_tempdir()
+  utils::unzip(f, exdir = d)
+
+  doc <- xml2::read_xml(file.path(d, "word", "document.xml"))
+  ns <- xml2::xml_ns(doc)
+  rows <- xml2::xml_find_all(doc, ".//w:tr", ns)
+
+  # Locate the spanner row and the column-label row by their text content.
+  row_text <- vapply(rows, xml2::xml_text, character(1))
```
llrs-roche:
```suggestion
  row_text <- xml2::xml_text(rows)
```

llrs-roche (approved):
Code looks good but I am not able to see any visual differences on tables on tests between main and this PR.  
A reproducible example with some output would make easier to review and verify the changes.

PR #279, opened 2026-06-30
annotate_riskdf() builds risk table at the plot.s x-axis breaks

### LINKED ISSUE #277: annotate_riskdf doesn't use updated axis breaks
When updating the breaks of axis of a ggplot the `annotate_riskdf()` uses the default of the plot not ones from the updated axis:

```r
# Preparing the Kaplan-Meier Plot
library(survival)
library("ggplot2")
use_lung <- lung
use_lung$arm <- factor(sample(c("A", "B", "C"), nrow(use_lung), replace = TRUE))
use_lung$status <- use_lung$status - 1 # Convert status to 0/1
use_lung <- na.omit(use_lung)

formula <- Surv(time, status) ~ arm
fit_kmg01 <- survfit(formula, use_lung)
surv_plot_data <- process_survfit(fit_kmg01)

plt_kmg01 <- gg_km(surv_plot_data) +
  scale_x_continuous(breaks = seq(from = 0, to = 1000, by = 100))

# Annotate Plot with Numbers at Risk Table
annotate_riskdf(plt_kmg01, fit_kmg01)
```

Image

Note how the table has values for 750 (correctly placed but with no label below) but not for 700 or 800.

### LINKED ISSUE #278: annotate_riskdf() ignores custom plot x-axis breaks for the risk table
## Summary

`annotate_riskdf()` does not honor custom x-axis breaks set on the plot. It always recomputes its own time points internally via `h_xticks()` (`labeling::extended(range(time), m = 5)`) and builds the "Numbers at Risk" table at those times. When a user sets study-specific breaks with `scale_x_continuous(breaks = ...)`, the risk table is computed at a different set of times than the plot ticks, so the two do not line up (e.g. the table shows a different number of columns than the plot has ticks).

## Reproducible example (synthetic data)

```r
library(crane)
library(survival)
library(ggplot2)

use_lung <- na.omit(transform(
  lung,
  status = status - 1,
  arm = factor(sample(c("A", "B"), nrow(lung), replace = TRUE))
))

fit <- survfit(Surv(time, status) ~ arm, use_lung)
spd <- process_survfit(fit)

my_breaks <- c(0, 100, 200, 300, 400, 500, 600, 700, 800)

plt <- gg_km(spd) +
  scale_x_continuous(breaks = my_breaks)

# the plot uses `my_breaks`, but the risk table is built at the auto grid:
crane:::h_xticks(data = broom::tidy(fit))
#> 0 250 500 750 1000

plt |> annotate_riskdf(fit)
# -> risk-table columns do not match the plot ticks
```

## Expected behavior

The risk table should use the same time points as the plot x-axis, so ticks and risk numbers align.

## Suggested fix

`annotate_riskdf()` already extracts the x-axis *positions* from the plot inside `df2gg_aligned()` (via `ggplot_build()`), so it should also take the *times* from the plot's resolved breaks rather than recomputing them with `h_xticks()`. Falling back to `h_xticks()` only when no usable numeric breaks are present keeps current behavior for plots that don't set custom breaks.

Melkiades:
* `annotate_riskdf()` now builds the "Numbers at Risk" table at the plot's x-axis breaks, so custom ticks set with `ggplot2::scale_x_continuous(breaks = ...)` are reflected in the table. (#278, @Melkiades)

A new internal helper `h_plot_xticks()` reads the resolved x-axis breaks from the built `ggplot` object. `annotate_riskdf()` uses these breaks for the risk-table timepoints, falling back to the previous automatic `h_xticks()` grid when the plot has no usable numeric breaks (e.g. a discrete x-axis).

**Reference GitHub issue associated with pull request.** closes #278

- [x] **All** GitHub Action workflows pass with a :white_check_mark:
- [x] Request a reviewer

- [x] Run `pkgdown::build_site()`. Check the R console for errors, and review the rendered website.
- [x] Code coverage is suitable for any new functions/features: `devtools::test_coverage()`

- [x] Update `NEWS.md` with the changes from this pull request under the heading "`# cards (development version)`". If there is an issue associated with the pull request, reference it in parentheses at the end update (see `NEWS.md` for examples).
- [x] **All** GitHub Action workflows pass with a :white_check_mark:
- [x] Approve Pull Request
- [x] Merge the PR. Please use "Squash and merge" or "Rebase and merge".

### DIFF
--- a/R/annotate_gg_km.R
+++ b/R/annotate_gg_km.R
@@ -69,6 +69,12 @@ NULL
 #'   object (not a combined `cowplot` object) because it requires exact X-axis
 #'   extraction.
 #'
+#' @details
+#' `annotate_riskdf()` reuses the plot's x-axis breaks, so custom ticks set with
+#' `ggplot2::scale_x_continuous(breaks = ...)` are reflected in the "Numbers at
+#' Risk" table. When the plot has no usable numeric breaks, the breaks are
+#' computed automatically.
+#'
 #' @return The function `annotate_riskdf` returns a `cowplot` object combining
 #'   the KM plot and the 'Numbers at Risk' table.
 #'
@@ -122,8 +128,19 @@ annotate_riskdf <- function(gg_plt,
   eargs <- utils::modifyList(default_eargs, list(...))
 
   # 2. Extract Data and Timepoints----------------------------------------------
+  # Use the plot's resolved x-axis breaks so the risk table lines up with the
+  # ticks the user set (e.g. via `scale_x_continuous(breaks = ...)`). A discrete
+  # x-axis has no time coordinates to align the risk table to, so error early
+  # with a clear message rather than failing further downstream.
   data <- broom::tidy(fit_km)
-  xticks <- h_xticks(data = data)
+  xticks <- .get_plot_xticks(gg_plt)
+  if (is.null(xticks)) {
+    rlang::abort(c(
+      "`gg_plt` must have a continuous (numeric) x-axis.",
+      "i" = "The risk table is aligned to survival times, which require numeric x-axis breaks.",
+      "x" = "A discrete or categorical x-axis was detected."
+    ))
+  }
   annot_tbl <- summary(fit_km, times = xticks, extend = TRUE)
 
   # 3. Format Strata Levels-----------------------------------------------------
--- a/R/gg_km_utils.R
+++ b/R/gg_km_utils.R
@@ -38,6 +38,32 @@ h_xticks <- function(data, xticks = NULL, max_time = NULL) {
   }
 }
 
+#' Extract X-axis Ticks From a Plot
+#'
+#' @description Reads the resolved x-axis breaks from a `ggplot` object so that
+#'   annotations (such as the risk table) can reuse the same tick positions the
+#'   plot displays. Uses the public `ggplot2::get_guide_data()` accessor rather
+#'   than the internal plot layout.
+#'
+#'   On a discrete axis `get_guide_data()` returns character values (the factor
+#'   levels), so a non-numeric `.value` signals a categorical axis with no
+#'   usable ticks.
+#'
+#' @param gg_plt (`ggplot`)\cr a Kaplan-Meier plot, typically from [gg_km()].
+#'
+#' @return A numeric vector of x-axis tick positions, or `NULL` when the plot has
+#'   no usable numeric breaks (e.g. a discrete x-axis).
+#'
+#' @keywords internal
+#' @noRd
+.get_plot_xticks <- function(gg_plt) {
+  guide <- ggplot2::get_guide_data(gg_plt, "x")
+  if (is.null(guide) || !is.numeric(guide$.value)) {
+    return(NULL)
+  }
+  guide$.value
+}
+
 #' @title Median Survival Summary Table
 #'
 #' @description Extracts and formats the median survival time and its confidence interval
--- a/tests/testthat/test-annotate_gg_km.R
+++ b/tests/testthat/test-annotate_gg_km.R
@@ -54,6 +54,45 @@ test_that("annotate_riskdf() handles all branches and inputs", {
   )
 })
 
+test_that("annotate_riskdf() uses the plot's custom x-axis breaks (#278)", {
+  my_breaks <- c(0, 100, 200, 300, 400, 500, 600, 700, 800)
+  p_breaks <- p_base + scale_x_continuous(breaks = my_breaks)
+
+  # the helper reads the resolved breaks back from the plot
+  expect_identical(
+    as.numeric(.get_plot_xticks(p_breaks)),
+    as.numeric(my_breaks)
+  )
+
+  # the risk table is built at those same times (one column per break)
+  annot_tbl <- summary(fit_strat, times = .get_plot_xticks(p_breaks), extend = TRUE)
+  expect_setequal(unique(annot_tbl$time), my_breaks)
+
+  expect_no_error(annotate_riskdf(p_breaks, fit_strat))
+})
+
+test_that(".get_plot_xticks() returns NULL on a discrete x-axis", {
+  # a discrete x-axis returns character values, not numeric breaks
+  p_disc <- ggplot(use_lung, aes(x = arm, y = status)) +
+    geom_col()
+  expect_null(.get_plot_xticks(p_disc))
+
+  # numeric-looking factor levels ("1", "2", "3") must also be treated as
+  # discrete, not mistaken for numeric breaks
+  use_lung$grp <- factor(sample(c("1", "2", "3"), nrow(use_lung), replace = TRUE))
+  p_num_lvls <- ggplot(use_lung, aes(x = grp, y = status)) +
+    geom_col()
+  expect_null(.get_plot_xticks(p_num_lvls))
+})
+
+test_that("annotate_riskdf() errors early on a discrete x-axis", {
+  # a KM axis is always continuous time; a categorical axis has no coordinates
+  # to align the risk table to, so it errors with an informative message
+  p_disc <- ggplot(use_lung, aes(x = arm, y = status)) +
+    geom_col()
+  expect_snapshot(annotate_riskdf(p_disc, fit_strat), error = TRUE)
+})
+
 # ------------------------------------------------------------------------------
 # 3. TESTS FOR annotate_surv_med()
 # ------------------------------------------------------------------------------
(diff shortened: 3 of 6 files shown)


Melkiades:
recheck

R/annotate_gg_km.R
```diff
@@ -64,7 +64,11 @@
 NULL
 
 #' @describeIn annotate_gg_km The function `annotate_riskdf` adds a "Numbers
-#'   at Risk" table below a Kaplan-Meier plot using `patchwork`.\cr
+#'   at Risk" table below a Kaplan-Meier plot using `patchwork`. The table reuses
+#'   the plot's x-axis breaks, so custom ticks set with
+#'   `ggplot2::scale_x_continuous(breaks = ...)` are reflected in the table; when
+#'   the plot has no usable numeric breaks the breaks are computed
+#'   automatically.\cr
```
Melkiades: Belongs to details

R/gg_km_utils.R
```diff
@@ -38,6 +38,39 @@ h_xticks <- function(data, xticks = NULL, max_time = NULL) {
...
+#'   `ggplot2`) are dropped.
+#'
+#' @param gg_plt (`ggplot`)\cr a Kaplan-Meier plot, typically from [gg_km()].
+#'
+#' @return A numeric vector of x-axis tick positions, or `NULL` when the plot has
+#'   no usable numeric breaks (e.g. a discrete x-axis).
+#'
+#' @keywords internal
+#' @examples
+#' NULL
```
Melkiades: nord

Melkiades: keeping keyword

R/gg_km_utils.R
```diff
@@ -38,6 +38,38 @@ h_xticks <- function(data, xticks = NULL, max_time = NULL) {
...
+#'
+#' @keywords internal
+#' @noRd
+.get_plot_xticks <- function(gg_plt) {
+  built <- tryCatch(ggplot2::ggplot_build(gg_plt), error = function(e) NULL)
+  if (is.null(built)) {
+    return(NULL)
+  }
+  params <- built$layout$panel_params[[1]]
+  breaks <- params$x$breaks
```
llrs-roche: I would use the [default recommended on extending ggplot2](https://ggplot2.tidyverse.org/articles/extending-ggplot2.html#creating-new-guides) as this relies on current implementation of the object (which might change):

```suggestion
  breaks <- ggplot2::get_guide_data(gg_plt, "x")$.value
```

This will simplify a lot the function as the plot as it doesn't need to be built

tests/testthat/test-annotate_gg_km.R
```diff
@@ -54,6 +54,30 @@ test_that("annotate_riskdf() handles all branches and inputs", {
...
+
+  expect_no_error(annotate_riskdf(p_breaks, fit_strat))
+})
+
+test_that(".get_plot_xticks() falls back to NULL without numeric breaks", {
+  # discrete x-axis has no usable numeric breaks
+  p_disc <- ggplot(use_lung, aes(x = arm, y = status)) +
+    geom_col()
+  expect_null(.get_plot_xticks(p_disc))
+})
```
llrs-roche: I think an extra test for `annotate_riskdf()` with categorical axis should also be added

llrs-roche (commented):
This is an elegant solution but I think the implementation can be simplified.

tests/testthat/test-annotate_gg_km.R
Melkiades: Good point

R/gg_km_utils.R
Melkiades: Done, switched to `get_guide_data()`. No reason to reach into `panel_params` when there is a public accessor.

One thing worth flagging: on a discrete axis `.value` comes back as the factor positions (1, 2, ...) not `NA`, so I kept a small guard that only keeps a break when its label parses back to the same number. Otherwise a categorical axis would sneak through as times 1 and 2. A KM axis is always continuous time anyway, so this is just there so the discrete case bails out cleanly.

ggplot2 is already pinned `>= 4.0.0` in DESCRIPTION so `get_guide_data()` is safe to rely on.

tests/testthat/test-annotate_gg_km.R
Melkiades: Added. Turns out a discrete x cannot actually produce a risk table (there are no time coordinates to align to), so `df2gg_aligned()` errors with `None of the table columns match the categorical X-axis levels in the plot.`. I snapshotted that instead of pretending it renders. `.get_plot_xticks()` returns `NULL` on the categorical axis as expected.

R/gg_km_utils.R
```diff
@@ -38,6 +38,40 @@ h_xticks <- function(data, xticks = NULL, max_time = NULL) {
...
+#' @keywords internal
+#' @noRd
+.get_plot_xticks <- function(gg_plt) {
+  guide <- tryCatch(ggplot2::get_guide_data(gg_plt, "x"), error = function(e) NULL)
+  if (is.null(guide) || is.null(guide$.value)) {
+    return(NULL)
+  }
+  breaks <- suppressWarnings(as.numeric(guide$.value))
+  labels <- suppressWarnings(as.numeric(guide$.label))
+  keep <- !is.na(breaks) & !is.na(labels) & breaks == labels
```
llrs-roche: I guess this is for categorical breaks. But won't be enough to check that `is.numeric(guide$.value)`?

R/gg_km_utils.R
```diff
@@ -38,6 +38,40 @@ h_xticks <- function(data, xticks = NULL, max_time = NULL) {
...
+#'
+#' @param gg_plt (`ggplot`)\cr a Kaplan-Meier plot, typically from [gg_km()].
+#'
+#' @return A numeric vector of x-axis tick positions, or `NULL` when the plot has
+#'   no usable numeric breaks (e.g. a discrete x-axis).
+#'
+#' @keywords internal
+#' @noRd
+.get_plot_xticks <- function(gg_plt) {
+  guide <- tryCatch(ggplot2::get_guide_data(gg_plt, "x"), error = function(e) NULL)
```
llrs-roche: Do we need to protect against failures here? It is only extracting data without even generating the plot

tests/testthat/test-annotate_gg_km.R
```diff
@@ -54,6 +54,40 @@ test_that("annotate_riskdf() handles all branches and inputs", {
...
+
+test_that("annotate_riskdf() errors on a discrete x-axis", {
+  # a KM axis is always continuous time; a categorical axis has no coordinates
+  # to align the risk table to, so the helper returns NULL and the downstream
+  # alignment errors with a clear message
+  p_disc <- ggplot(use_lung, aes(x = arm, y = status)) +
+    geom_col()
+  expect_null(.get_plot_xticks(p_disc))
+  expect_snapshot(annotate_riskdf(p_disc, fit_strat), error = TRUE)
```
llrs-roche, line 93: The validation team warns against snapshots.
Can we `expect_error()` with an informative error message (might require modifying `annotate_riskdf()`)

tests/testthat/test-annotate_gg_km.R
llrs-roche: I guessed so, but I think the function should error earlier and provide informative error message to the users.
Just in case someone tries to use it with category baseline, T1, T2 and T3 or something similar.

llrs-roche (commented):
Thanks for the updates. Some more comments that will be quick to resolve.

tests/testthat/test-annotate_gg_km.R
Melkiades: I started using the above when Daniel joined, as he suggested that for error. To be honest, I find snapshotting errors extremely more readable and close to the user feedback. Many times I had to manually copy paste specific errors which is just a loss of time if you change the text of an error. I do not know why they do not suggest it, but gtsummary, cards, cardx and crane have it, and I would keep it. For me, I would even encourage its use tbh

R/gg_km_utils.R
Melkiades: Yes, much cleaner. On a discrete axis `.value` comes back as character (the factor levels), so `is.numeric(guide$.value)` is enough and it also covers the numeric-looking levels case ("1","2","3") you mention below. Switched to it and dropped the label parsing. Added a test with numeric-looking factor levels too.

R/gg_km_utils.R
Melkiades: Good point, removed the `tryCatch`. `get_guide_data()` just reads the resolved guide, no build step that could blow up here.

tests/testthat/test-annotate_gg_km.R
Melkiades: Done, moved the check into `annotate_riskdf()` so it errors up front:

```
! `gg_plt` must have a continuous (numeric) x-axis.
i The risk table is aligned to survival times, which require numeric x-axis breaks.
x A discrete or categorical x-axis was detected.
```

This now catches the baseline/T1/T2/T3 case you were worried about before it reaches the alignment code.

llrs-roche (approved):
LGTM and on the hands of users

PR #283, opened 2026-07-02
Add modify_split_caption() to label paginated splits and hide the split column (#282)

### LINKED ISSUE #282: Add modify_split_caption() to label paginated splits and hide the split column
When paginating a listing with `gtsummary::tbl_split_by_rows(variable_level = <col>)` (e.g. splitting a lab listing by `PARAM`, or a demographics listing by treatment), two manual post-processing steps are always needed:

1. The split value is stored in `attr(tbl, "variable_level")` as a bare level (e.g. `"WEIGHT"`). Decorators (citril) inject it as the page subtitle, but it usually needs a prefix, e.g. `"Parameter: WEIGHT"`.
2. The split column is redundant on each single-value page and should be hidden.

Currently this is repeated inline in every paginated template:

```r
gts |>
  purrr::map(\(tbl) {
    lvl <- attr(tbl, "variable_level")
    tbl <- gtsummary::modify_column_hide(tbl, columns = "PARAM")
    attr(tbl, "variable_level") <- paste0("Parameter: ", lvl)
    tbl
  }) |>
  structure(class = class(gts))
```

### Proposal

Add `modify_split_caption()` to `R/tbl_listing.R` (sibling of `remove_duplicate_keys()` / `add_blank_rows()`, shared `@rdname tbl_listing`):

```r
modify_split_caption(
  x,
  spl_col,
  pattern = "Parameter: {spl_level}",  # glue; {spl_level} is the split value
  hide_spl_col = TRUE                   # hidden silently; no-op if col absent/already hidden
)
```

- Rewrites each split page's `variable_level` label via the glue `pattern` (default suits BDS/lab tables split by `PARAM`).
- Hides `spl_col` on every page by default; silent when the column is absent or already hidden.
- Works on a single `tbl_split` element or the whole list.
- No citril/internal references — operates only on the gtsummary-native `variable_level` attribute, so crane stays open-source clean.

Melkiades:
* Added `modify_split_caption()` to label each page of a split listing from its split level (via a glue `pattern`, default `"Parameter: {spl_level}"`) and hide the now-redundant split column. (#282, @Melkiades)

When a listing is paginated with `gtsummary::tbl_split_by_rows(variable_level = <col>)` (e.g. a lab listing split by `PARAM`, or a demographics listing split by treatment), two post-processing steps were previously repeated inline in every template: prefixing the split level for the page caption/subtitle, and hiding the redundant split column. `modify_split_caption()` bundles both:

```r
tbl_listing(df, split_by_rows = list(variable_level = "PARAM")) |>
  modify_split_caption(spl_col = "PARAM")
# -> each page caption "Parameter: <level>"; PARAM column hidden
```

- `pattern` is a single glue string (`{spl_level}` is the split value); default suits BDS/lab tables.
- `hide_spl_col = TRUE` by default; silent when the column is absent or already hidden.
- Works on a single split element or the whole `tbl_split` list; row-number splits (no `variable_level`) are skipped.
- Keeps the gtsummary-native `variable_level` attribute in sync with the caption, so decorators that read it (e.g. citril) show the same text — no external dependency added.

Co-located in `R/tbl_listing.R` under `@rdname tbl_listing`, next to `remove_duplicate_keys()`.

**Reference GitHub issue associated with pull request.** closes #282

- [ ] Request a reviewer

- [x] Code coverage is suitable for any new functions/features: `devtools::test_coverage()`

- [x] Update `NEWS.md` with the changes from this pull request under the heading "`# crane (development version)`". If there is an issue associated with the pull request, reference it in parentheses at the end update (see `NEWS.md` for examples).
- [ ] Approve Pull Request


R/modify_split_caption.R
```diff
@@ -0,0 +1,104 @@
...
+    )
+  }
+
+  # checks ---------------------------------------------------------------------
+  check_not_missing(x)
+  check_class(x, "gtsummary")
+  check_not_missing(spl_col)
+  check_string(spl_col)
+  check_string(pattern)
+  check_scalar_logical(hide_spl_col)
```
llrs-roche: Checks/assertions should go before the map over a list of split tables as it is used there

R/modify_split_caption.R
```diff
@@ -0,0 +1,104 @@
...
+
+  # map over a list of split tables --------------------------------------------
+  if (is.list(x) && inherits(x[[1]], "gtsummary")) {
+    return(
+      map(
+        x,
+        modify_split_caption,
+        spl_col = spl_col,
+        pattern = pattern,
+        hide_spl_col = hide_spl_col
```
llrs-roche, line 80: Consider this to make it more explicit that we are calling the function itself. (Not tested, but I think it should work)

```suggestion
        Recall,
        spl_col = spl_col,
        pattern = pattern,
        hide_spl_col = hide_spl_col
```

R/modify_split_caption.R
```diff
@@ -0,0 +1,104 @@
...
+  spl_level <- attr(x, "variable_level")
+  if (!is_empty(spl_level)) {
+    caption <- glue::glue(pattern)
+    x <- gtsummary::modify_caption(x, as.character(caption))
+    # keep the attribute in sync so decorators that read it show the same text
+    attr(x, "variable_level") <- as.character(caption)
+  }
+
+  # hide the split column, if present and not already hidden --------------------
+  if (isTRUE(hide_spl_col) && spl_col %in% x$table_styling$header$column) {
```
llrs-roche: The code below checks for all_of but on the condition it doesn't. But it assumes a single logical value:
```suggestion
  if (isTRUE(hide_spl_col) && all(spl_col %in% x$table_styling$header$column)) {
```

tests/testthat/test-modify_split_caption.R
```diff
@@ -0,0 +1,95 @@
...
+      modify_split_caption(spl_col = "trt")
+  )
+
+  # one page per treatment level, class preserved
+  expect_s3_class(out, "tbl_split")
+  expect_s3_class(out[[1]], "tbl_listing")
+
+  # default pattern is applied and kept in sync with the variable_level attribute
+  expect_equal(as.character(out[[1]]$table_styling$caption), "Parameter: Drug B")
+  expect_equal(attr(out[[1]], "variable_level"), "Parameter: Drug B")
```
llrs-roche, line 24: Maybe test on other pages? To match the test with the actual tests performed

R/modify_split_caption.R
```diff
@@ -0,0 +1,104 @@
+#' Caption and Trim Split Tables
+#'
+#' @description
+#' Post-process a table that was paginated with
+#' [gtsummary::tbl_split_by_rows()] (via its `variable_level` argument). For each
+#' page, a caption is built from the split level using a [glue][glue::glue]
```
llrs-roche: Maybe this is enough:

```suggestion
#' page, a caption is built from the split level using a [glue()]
```

R/modify_split_caption.R
```diff
@@ -0,0 +1,104 @@
...
+#'
+#' # Split a listing by treatment, then caption each page and drop the column
+#' split_tbls <- tbl_listing(df, split_by_rows = list(variable_level = "trt")) |>
+#'   modify_split_caption(spl_col = "trt", pattern = "Treatment: {spl_level}")
+#' split_tbls[[1]]
+#'
+#' # Keep the split column and use the default "Parameter:" caption
+#' tbl_listing(df, split_by_rows = list(variable_level = "trt")) |>
+#'   modify_split_caption(spl_col = "trt", hide_spl_col = FALSE) |>
+#'   _[[1]]
```
llrs-roche: I thought this  `_[[1]]` wouldn't work! Nice!!

R/modify_split_caption.R
```diff
@@ -0,0 +1,104 @@
...
+#'   dplyr::select(trt, age, grade) |>
+#'   dplyr::arrange(trt)
+#'
+#' # Split a listing by treatment, then caption each page and drop the column
+#' split_tbls <- tbl_listing(df, split_by_rows = list(variable_level = "trt")) |>
+#'   modify_split_caption(spl_col = "trt", pattern = "Treatment: {spl_level}")
+#' split_tbls[[1]]
+#'
+#' # Keep the split column and use the default "Parameter:" caption
+#' tbl_listing(df, split_by_rows = list(variable_level = "trt")) |>
```
llrs-roche: Some suggestions to make the example shorter and focused on the new function

```suggestion
#' lst <- trial |>
#'   dplyr::select(trt, age, grade) |>
#'   dplyr::arrange(trt) |>
#'   tbl_listing(df, split_by_rows = list(variable_level = "trt"))
#' 
#' # Split a listing by treatment, then caption each page and drop the column
#' modify_split_caption(lst, spl_col = "trt", pattern = "Treatment: {spl_level}")[[1]]
#'
#' # Keep the split column and use the default "Parameter:" caption
#' lst |>
```

llrs-roche (commented):
Not sure if expected but on Rstudio when visualizing this split I see this:

Image

What it is surprising is that the pattern is at the center of the table. I'm not sure how it is on other formats but might be worth checking

R/modify_split_caption.R
Melkiades: Done. Moved the argument checks above the map so a list of split tables is validated once, up front. The `check_class(x, "gtsummary")` stays after since it must not fire on the list itself.

R/modify_split_caption.R
Melkiades: Good point, done.

tests/testthat/test-modify_split_caption.R
Melkiades: Added. Now asserts caption, `variable_level` and the hide flag on both pages (Drug B and Drug A).

R/modify_split_caption.R
Melkiades: I tried it but `Recall` doesn't work inside `map()` here -- it resolves to the internal `map_()` caller, not `modify_split_caption`, so you get `argument ".x" is missing, with no default`. Kept the explicit function name.

R/modify_split_caption.R
Melkiades: Done.

R/modify_split_caption.R
Melkiades: Yeah the placeholder `_[[1]]` is handy. It's gone now anyway after shortening the examples below.

R/modify_split_caption.R
Melkiades: Done. Build `lst` once then two focused calls. Small tweak vs your suggestion: `tbl_listing()` takes the data as first arg, so after piping `df` in I dropped the extra `df` (`tbl_listing(df, ...)` would pass it twice).

Melkiades:
On the centered caption: that's the renderer's default, not something this function sets -- we just call `gtsummary::modify_caption()` and the alignment comes from how each backend prints the caption. Worth checking how it lands in docx/flextable (the actual target format) separately; I'd rather not special-case alignment inside `modify_split_caption()`.

llrs-roche (approved):

PR #289, opened 2026-07-23
Add pk helper functions

qiliu1013:
**What changes are proposed in this pull request?**
* Style this entry in a way that can be copied directly into `NEWS.md`. (#<issue number>, @<username>)

Provide more detail here as needed.

Add cv(), geom_cv(), geom_mean(), fmt_3sig() and impuation_rules for the PK TLGs.

Close #287 

--------------------------------------------------------------------------------

Pre-review Checklist (if item does not apply, mark is as complete)
- [ ] **All** GitHub Action workflows pass with a :white_check_mark:
- [ ] PR branch has pulled the most recent updates from master branch: `usethis::pr_merge_main()`
- [ ] If a bug was fixed, a unit test was added.
- [ ] Code coverage is suitable for any new functions/features (generally, 100% coverage for new code): `devtools::test_coverage()`
- [ ] Request a reviewer

Reviewer Checklist (if item does not apply, mark is as complete)

- [ ] If a bug was fixed, a unit test was added.
- [ ] Run `pkgdown::build_site()`. Check the R console for errors, and review the rendered website.
- [ ] Code coverage is suitable for any new functions/features: `devtools::test_coverage()`

When the branch is ready to be merged:
- [ ] Update `NEWS.md` with the changes from this pull request under the heading "`# cards (development version)`". If there is an issue associated with the pull request, reference it in parentheses at the end update (see `NEWS.md` for examples).
- [ ] **All** GitHub Action workflows pass with a :white_check_mark:
- [ ] Approve Pull Request
- [ ] Merge the PR. Please use "Squash and merge" or "Rebase and merge".

PR #294, opened 2026-08-10
Fix PK helper defects and add tests

### LINKED ISSUE #293: PK helpers: remove dead postdose arg, fix geom_mean, vectorize formatters, add tests
## Summary

The PK helper functions added in #289 (`R/pk_helper.R`) have a few defects and gaps that should be addressed.

### 1. `imputation_rules()` — `postdose` argument is dead code
The two `1/3` rule branches (`!postdose` and `postdose`) are byte-for-byte identical, so the `postdose` flag has no effect on the output. The argument adds a required parameter and implies behaviour that does not exist.

Post-dose vs pre-dose handling is a **data step** (e.g. substituting post-dose LTRs with half the LLOQ before summarizing), not a display decision. That logic belongs upstream in the caller/template, before statistics are computed. `imputation_rules()` should only decide how an already-computed statistic is *displayed* given the BLQ ratio.

**Proposed fix:** remove `postdose` from the signature and collapse the two identical branches into one.

### 2. `geom_mean()` returns `NaN` for all-`NA` input
`geom_mean(c(NA, NA))` returns `NaN` rather than `NA_real_`. The `any(x <= 0, na.rm = na.rm)` guard can also propagate `NA` into `if()` when `na.rm = FALSE`.

**Proposed fix:** drop `NA`s up front (when `na.rm = TRUE`), return `NA_real_` for the empty/all-non-positive cases.

### 3. Formatters are not vectorized and bake in domain logic
`fmt_3sig()` and `fmt_pct()` accept only length-1 input and return the string `"NE"` for `NA`. "Not estimable" is analysis semantics, not formatting — mixing it into a generic formatter is surprising and makes the functions hard to reuse.

**Proposed fix:** vectorize both formatters and return `NA_character_` for `NA`/non-finite input. The `"NE"`/`"ND"` substitution stays in `imputation_rules()` where it belongs.

### 4. Missing documentation
`cv()`, `geom_cv()`, and `fmt_pct()` lack `@description` and lifecycle badges that the other helpers have.

### 5. No test coverage
`R/pk_helper.R` shipped without a `tests/testthat/test-pk_helper.R`.

## Scope
All of the above will be fixed in one PR. The public API change is the removal of `postdose` from `imputation_rules()`; the dependent `tlg.templates` PKCT01 template call will be updated separately.

Melkiades:
* `imputation_rules()` drops the `postdose` argument, which had no effect (the pre-dose and post-dose branches were identical). Timing-dependent BLQ substitution belongs in the data before summarizing. (#293)
* `geom_mean()` returns `NA_real_` instead of `NaN` when all values are missing. (#293)
* `fmt_3sig()` and `fmt_pct()` are now vectorized and return `NA_character_` for `NA` or non-finite input. (#293)

Follow-up to #289, which added the PK helpers without tests. This PR fixes the defects found on review and adds full test coverage for `R/pk_helper.R`.

The only public API change is the removal of `postdose` from `imputation_rules()`. The dependent `tlg.templates` PKCT01 template call will be updated separately.

**Reference GitHub issue associated with pull request.** closes #293

- [x] **All** GitHub Action workflows pass with a :white_check_mark:
- [ ] Request a reviewer

- [x] Update `NEWS.md` with the changes from this pull request under the heading "`# cards (development version)`". If there is an issue associated with the pull request, reference it in parentheses at the end update (see `NEWS.md` for examples).
- [ ] Approve Pull Request


R/pk_helper.R
```diff
@@ -36,109 +37,118 @@ geom_cv <- function(x) {
...
+#' be applied upstream, before the statistics are computed.
+#'
 #' @examples
-#' imputation_rules("1.23", "Mean", blq_ratio = 0.5, postdose = TRUE, rule = "1/3")
+#' imputation_rules("1.23", "Mean", blq_ratio = 0.5, rule = "1/3")
 #'
 #' @export
-imputation_rules <- function(stat_val, label, blq_ratio, postdose, rule = "1/3") {
-  if (is.na(blq_ratio)) {
+imputation_rules <- function(stat_val, label, blq_ratio, rule = "1/3") {
```
Melkiades: @martincadek note: imputation rules can be missing values. Add pk specifics for this imputation rules (e.g. pk_*)

R/pk_helper.R
```diff
@@ -36,109 +37,118 @@ geom_cv <- function(x) {
...
     ),
 
-    # 1/3 rule: pre-dose
-    rule == "1/3" & !postdose & blq_ratio > 1 / 3 ~ dplyr::case_when(
-      label %in% c("Median", "Max", "Geom_mean", "No. obs.", "Number of LTR/BLQ") ~ stat_val,
-      TRUE ~ "ND"
-    ),
-
-    # 1/3 rule: post-dose
-    rule == "1/3" & postdose & blq_ratio > 1 / 3 ~ dplyr::case_when(
```
Melkiades, line 141: keep postdose handling as PK SME intended. Just to flag

R/pk_helper.R
```diff
@@ -36,109 +37,118 @@ geom_cv <- function(x) {
...
-    is.null(rule) ~ stat_val,
 
     # 1/2 rule
     rule == "1/2" & blq_ratio > 0.5 ~ dplyr::case_when(
       label %in% c("Max", "No. obs.", "Number of LTR/BLQ") ~ stat_val,
       TRUE ~ "ND"
     ),
 
-    # 1/3 rule: pre-dose
-    rule == "1/3" & !postdose & blq_ratio > 1 / 3 ~ dplyr::case_when(
```
Melkiades, line 135: Improve readability

shajoezhu (approved):
lgtm! thanks @Melkiades

R/pk_helper.R
```diff
@@ -109,49 +109,54 @@ fmt_pct <- function(x) {
...
-#' Any substitution of BLQ values in the data (e.g. based on dosing timing) must
-#' be applied upstream, before the statistics are computed.
-#'
 #' @examples
-#' imputation_rules("1.23", "Mean", blq_ratio = 0.5, rule = "1/3")
+#' imputation_rules("1.23", "Mean", blq_ratio = 0.5, postdose = TRUE, rule = "1/3")
 #'
 #' @export
-imputation_rules <- function(stat_val, label, blq_ratio, rule = "1/3") {
+imputation_rules <- function(stat_val, label, blq_ratio, postdose, rule = "1/3") {
```
Melkiades: pk_imputation_rules

R/pk_helper.R
```diff
@@ -109,49 +109,54 @@ fmt_pct <- function(x) {
...
   if (is.null(rule) || is.na(blq_ratio)) {
     return(stat_val)
   }
   force(label)
+  force(postdose)
```
Melkiades: align the checks to the rest of the package

NEWS.md
```diff
@@ -1,3 +1,9 @@
+# crane 0.3.3.9008
+
+* `geom_mean()` returns `NA_real_` instead of `NaN` when all values are missing. (#293)
+
+* `fmt_3sig()` and `fmt_pct()` are now vectorized and return `NA_character_` for `NA` or non-finite input. (#293)
+
```
Melkiades: make it better and related to pk outputs. where is pk_imputation_rule?

R/pk_helper.R
```diff
@@ -36,74 +37,90 @@ geom_cv <- function(x) {
...
+#' @returns A `character` vector the same length as `x`, with `NA_character_`
+#'   where `x` is `NA` or non-finite.
 #'
 #' @examples
 #' fmt_3sig(0.001234)
 #' fmt_3sig(123456)
+#' fmt_3sig(c(0.001234, NA, 123456))
 #'
 #' @export
 fmt_3sig <- function(x) {
```
Melkiades, line 102: add checks to all exported functions

R/pk_helper.R
```diff
@@ -36,74 +37,90 @@ geom_cv <- function(x) {
...
 #'
-#' @returns A single numeric value, or `NA_real_` if any non-`NA` value is <= 0.
+#' @returns A single numeric value, or `NA_real_` if there are no usable values
+#'   or if any non-`NA` value is <= 0.
 #'
 #' @examples
 #' geom_mean(c(1, 2, 3, 4, 5))
 #'
 #' @export
 geom_mean <- function(x, na.rm = TRUE) {
```
Melkiades, line 80: Do we need to write this manually? isnt there some option for this? or is it specific to pk?

shajoezhu: pk specific for now. they have their computation, let's get it work first, and we will come back for refacoring if needs

Melkiades: I would change the name into pk_* as it is specific to these calculations

shajoezhu: dont worry about it, it was almost direct copy from tern, they are using a mix of nest 1.0 and 2.0. this is too detailed change i think

Melkiades: I just rework the docs so it is part of pk stuff

Kcdahn:
Can we put fmt_3sig() and fmt_pct()  back to showing "NE" when those stats are null/na?  We added the condition on purpose as when n=1,  SD, CV% percent's are not estimable so added that to match our Stream and Nest 1.0  versions of this table ( that place "NE" in those cell in these cases).   We simulated a NFRLT=999 with one record to produce and test this feature. 

Target output:
image

How it is in the new crane helper:

image

R/pk_param.R
```diff
@@ -1,14 +1,18 @@
-#' Generate PK reference dataset
+#' @rdname pk_helpers
 #'
```
Melkiades: join the files at this point

Melkiades:
Thanks @Kcdahn — the `"NE"` for non-estimable stats (SD, CV%, geometric CV% when `n = 1`) is preserved, just relocated so the output matches your Stream / Nest 1.0 target.

Rather than have `fmt_3sig()`/`fmt_pct()` emit `"NE"` (they're generic numeric formatters, also used outside PK), the substitution now lives in `pk_imputation_rules()`, which the table already calls per statistic. The rule is: any missing statistic becomes `"NE"`, except counts (`No. obs.`, `Number of LTR/BLQ`), which are always available. At `n = 1` that's exactly SD / CV% / geometric CV% / geometric mean → `"NE"`, so your simulated `NFRLT=999` single-record case renders identically to the target image.

Net effect for the table: unchanged — the `NFRLT=999` row shows `"NE"` in those cells. The only difference is *where* the logic sits (display layer, one place) so the formatters stay purely numeric. Could you re-run your `NFRLT=999` check against this branch to confirm it matches?

R/pk_param.R
Melkiades: Done — merged `R/pk_param.R` into `R/pk_helper.R` (both already share the `?pk_helpers` page) and deleted the separate file. The 398-row reference table is preserved verbatim; `pk_helpers.Rd` now sources from the single file. Pushed in `aa7a3b1`.

mhallal1:
@qiliu1013 please have a look at this MR before merge.

PR #308, opened 2026-08-19
Add se to annotate_lineplot_df() summary table

### LINKED ISSUE #307: Add SE to annotate_lineplot_df() summary table
`gg_lineplot()` already draws Mean +/- 1xSE error bars via `variability = "se"`, but the summary table produced by `annotate_lineplot_df()` cannot display an SE row. Its `summary_stats` only supports `n`, `mean`, `sd`, `median`, and `iqr`.

Add `"se"` as a `summary_stats` option so the table below the plot can report the standard error alongside (or instead of) SD. Computed as `sd / sqrt(n)` via the existing cards glue engine (`{se}`).

Melkiades:
## What changed

* `annotate_lineplot_df()` gains `"se"` as a `summary_stats` option, reporting the standard error (`sd / sqrt(n)`) in the summary table below the plot. (#307)

## Why

`gg_lineplot()` already draws Mean +/- 1xSE error bars via `variability = "se"`, but the summary table under the plot could only report `n`, `mean`, `sd`, `median`, and `iqr`. This adds the missing SE row so the table can match the error bars.

SE is computed through the existing cards glue engine (`{se}`), no new dependency. Verified numerically against `sd / sqrt(n)` in a new test.

Closes #307


R/annotate_gg.R
```diff
@@ -72,6 +77,21 @@ annotate_lineplot_df <- function(gg_plt,
...
+    x <- stats::na.omit(x)
+    n <- length(x)
+    mean(x) - stats::qt(.ci_alpha, df = max(1, n - 1)) * stats::sd(x) / sqrt(n)
+  }
+  conf.high <- function(x) {
+    x <- stats::na.omit(x)
+    n <- length(x)
+    mean(x) + stats::qt(.ci_alpha, df = max(1, n - 1)) * stats::sd(x) / sqrt(n)
+  }
+
```
Melkiades: cant this be more explicit? or should it be modifiable?

Melkiades: Isnt there a general function that can be used?

mhallal1 (approved):
tested and reviewed, works as expected.

